PR #38, opened 2024-06-17
Feature/27 coding standards

### LINKED ISSUE #27: Draft, document, and integrate coding standards
As our team grows, it's necessary to follow specific coding standards. Let's start and iterate together.

- [x] Check on official C# + Unity formatting guidelines
- [x] Check for other Unity projects and what standards they followed
- [ ] Document our initially drafted standards (and/or link to official ones)
- [ ] Integrate these standards into VS and Rider (PoC)
- [ ] Discuss first iteration with team and iterate

Out of scope:
* Changing existing code
* Full blown solution of everything - think of it more like a vertical slice. Especially testing the integration within IDEs. And do not overcomplicate it by adding hundreds of rules

JaXt0r:
# Please review briefly:

* Start Rider and check if the following lines will hint a yellow warning:
* (optional) Test the same warnings with VS22
---
- [ ] 1. ```private bool isTest;``` <-- Needs to be underscored
image
---
- [ ] 2. ```protected bool isTest2;``` <-- Needs to be PascalCase like every non-private field
image
---
- [ ] 3. Braces are needed.
```c#
if (isActiveAndEnabled)
    return;
```
image
image

---
# Out-of-scope:
* I added only ~5 checks. Please look at .editorconfig for details. We can update in the future whenever needed.
* I only linked to "auto-format on save" documentation. I think this is sufficient.

### DIFF
--- a/Docs/development/style-guide.md
+++ b/Docs/development/style-guide.md
@@ -1,0 +1,270 @@
+```c#
+// STYLE SHEET EXAMPLE
+// Taken and slighly altered from Thomas Krogh-Jacobsen (Team lead, Tech Content Marketing Management at Unity Technologies)
+// @see: https://github.com/thomasjacobsen-unity/Unity-Code-Style-Guide/blob/master/StyleExample.cs
+
+//-------------------------
+
+// NAMING/CASING:
+// - Use Pascal case (e.g. ExamplePlayerController, MaxHealth, etc.) unless noted otherwise
+// - Use camel case (e.g. examplePlayerController, maxHealth, etc.) for local variables, parameters.
+// - Avoid snake case, kebab case, Hungarian notation
+// - The source file name must match. 
+
+// FORMATTING:
+// - Use Allman (opening curly braces on a new line).
+// - Keep lines short. Consider horizontal whitespace. Use a standard line width in your style guide 120 characters. 
+// - Use a single space before flow control conditions, e.g. while (x == y)
+// - Avoid spaces inside brackets, e.g. x = dataArray[index]
+// - Use a single space after a comma between function arguments.
+// - Don’t add a space after the parenthesis and function arguments, e.g. CollectItem(myObject, 0, 1);
+// - Don’t use spaces between a function name and parenthesis, e.g. DropPowerUp(myPrefab, 0, 1);
+// - Use vertical spacing (extra blank line) for visual separation. 
+
+// COMMENTS:
+// - Rather than simply answering "what" or "how," comments can fill in the gaps and tell us "why."
+// - Use the // comment to keep the explanation next to the logic.
+// - Use a Tooltip instead of a comment for serialized fields. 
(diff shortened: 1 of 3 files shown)


ThomasMountainborn:
> protected bool isTest2; <-- Needs to be PascalCase like every non-private field

Hmm, I find that a bit weird, since then you still can't immediately tell the difference between a local variable and a member. The difference between private and protected isn't very relevant in the code, but the difference between local and member is. I find it clearest this way:

PublicMember
_privateMember
localVariable

JaXt0r:
> > protected bool isTest2; <-- Needs to be PascalCase like every non-private field
> 
> Hmm, I find that a bit weird, since then you still can't immediately tell the difference between a local variable and a member. The difference between private and protected isn't very relevant in the code, but the difference between local and member is. I find it clearest this way:
> 
> PublicMember _privateMember localVariable

Yep. 100%. I didn't phrase it properly. But meant the same outcome as you said:

public PascalCase;
protected PascalCase;
private _underscoreCase;
Local: camelCase;

Docs/development/style-guide.md
```diff
@@ -0,0 +1,286 @@
...
+
+        // This groups data from the custom PlayerStats class in the Inspector.
+        [SerializeField] private PlayerStats _stats;
+
+        // This limits the values to a Range and creates a slider in the Inspector.
+        [Range(0f, 1f)] [SerializeField] private float _rangedStat;
+
+        // A tooltip can replace a comment on a serialized field and do double duty.
+        [Tooltip("This is another statistic for the player.")]
+        [SerializeField] private float _anotherStat;
```
lmichaelis: We should mention Unity-style for inspector fields here (camelCase). Do we adhere to it or do we ignore it and go with `PascalCase` anyways?

lmichaelis (commented):
Looks generally good :)

Docs/development/style-guide.md
JaXt0r: What do you mean? The example above leverages UnityFields the same as non-Unity fields. What is default and what would change in here?
```[SerializeField] private float _anotherStat;```

Comment (lmichaelis) [reply Docs/development/style-guide.md]: Ah did I misunderstand? Are we going to use private members even for properties set through the inspector?

Comment (JaXt0r) [reply Docs/development/style-guide.md]: I didn't include this as a recommendation. What's your recommendation? Then I can add this information.

Comment (JaXt0r) [reply Docs/development/style-guide.md]: Ok. I added another hint at this. Anything else needed? If not, feel free to accept the PR.

```c#
        // Use [SerializeField] attribute if you want to display a private field in Inspector.
        // Try to use private Fields for Components/MonoBehaviours, if you don't need them to be public.
        // Booleans ask a question that can be answered true or false.
        [SerializeField] private bool _isPlayerDead;
```

lmichaelis (approved):

PR #39, opened 2024-06-17
Refactor Global State

JaXt0r:
Draft PR to track issues or change requests for new GUZ architecture.


JaXt0r:
Hi Luis. Some findings:

**GameConfiguration.cs**
* I think it should reflect the production values we use for build. This makes it super easy to recreate the asset if there's a new field.
* Please let us store the prod-one into the Resource folder by default. I think of: Assets/Gothic-UnZENity-Core/Resources/GameConfigurations/GUZ-Features-Prod
* Changes to the GameConfiguration.cs to reflect prod values:
    * showMeshCullingGizmos = false
    * StartHour + StartMinute=08:00 <-- I'd prefer the official G1 start time. Makes it easier to debug "New Game" issues in the future.
    * sunLightColor = ... <-- I think it was a different color in FeatureFlags. Some yellow-ish one?
    * spawn specific Npcs <-- Field is missing in your Scriptable object.
    * enableDecalVisuals/enableParticleEffects = false <-- These are experimental / WIP features. Maybe we highlight those in a separate group?
* I like the new grouping based on Developer/Audio/Lighting/...

JaXt0r:
**Bug: Barrier is rendered inside main menu (including thunder sound).**
(Hint: If Bugs like these aren't related to you and occur in main already, just name it here and create a separate bug report, please.)
image

Tested on main: No barrier in main menu.

JaXt0r:
**Bug: Oldmine (world switchings) isn't loading**. I tested with main where it works properly. In this branch, the world meshes aren't rendered.

But the world chunks are there... Strange:
image

JaXt0r:
**Bug: NPCs spawn one over another**
In main it's the case, whenever we set:
* Spawn NPCs=true
* EnableNpcRoutines=false

image

But in my case, I activated both:



---

I found the source: It's because inside NpcCreator.cs you still reference the FeatureFlag instead of the new GameConfiguration asset.
image

JaXt0r:
**Hint: All your managers have an Init() function.** How about naming an interface to ensure proper use? The interface method can also centrally describe _why_ it's needed.

Or even better: Why not calling this function directly at construct()?  
A I see... It seems we need to wait for Start() to initialize properly.  Then back to my interface idea. ;-)

lmichaelis:
Ready for review. Please test: Everything :smile: 

Most important bits:
- [ ] Game starts?
- [ ] Lab starts and works?
- [ ] Game nenu works?
- [ ] Lighting correct?
- [ ] Sky okay?
- [ ] Barrier works?
- [ ] XR interactions okay?
- [ ] Configuration (via `GameConfiguration`) works?
- [ ] Worlds other than `world.zen` work?
- [ ] Transitions between worlds work?
- [ ] Works on XR hardware?
- [ ] Release works and build runs?

JaXt0r (commented):
Test

Assets/Gothic-UnZENity-Core/Scripts/ResourceLoader.cs
```diff
@@ -0,0 +1,241 @@
...
+		[CanBeNull]
+		public static Segment TryGetSegment([NotNull] string key)
+		{
+			// NOTE(lmichaelis): There is no caching required here, since the loader
+			//                   already caches segments upon loading them
+			return DmLoader.GetSegment(key);
+		}
+		
+		[CanBeNull]
+		public static DaedalusVm TryGetDaedalusVm([NotNull] string key)
```
JaXt0r, line 177: Is it really good to _not_ cache this call? Then wherever we call it from scritps, we will always reinitialize the VM, right?
I found this (e.g.) inside LabMusicHandler.Bootrstrap(). For Lab this is fine one time, but I expect we will see this usage and reinitialize wrong wherever we need data from other VMs again.

JaXt0r, line 177: Another question is also: Do we want to prepare the names of the "default" VMs from G1 on top? Not as Enum, but at least constants, where we can see the correct keys?

Assets/Gothic-UnZENity-Lab/Scripts/Handler/LabMusicHandler.cs
```diff
@@ -16,25 +18,18 @@ public class LabMusicHandler : MonoBehaviour, ILabHandler
 
         public void Bootstrap()
         {
-            var prototype = GameData.MusicVm.GetSymbolByName("C_MUSICTHEME_DEF");
+            var vm = ResourceLoader.TryGetDaedalusVm("MUSIC");
```
JaXt0r, line 21: We load the music VM uncached. I'd say it's better we load all the VMs during boot time. For normal game, we need them. For lab, this is no issue in terms of a few seconds of longer scene startup time.
(@see see my comment in ResourceLoader itself)

Assets/Gothic-UnZENity-Core/Scripts/Globals/GameData.cs
```diff
@@ -16,13 +16,9 @@ namespace GUZ.Core.Globals
 {
     public static class GameData
     {
-        public static Vfs Vfs;
         public static DaedalusVm GothicVm;
         public static DaedalusVm SfxVm; // Sound FX
         public static DaedalusVm PfxVm; // Particle FX
-        public static DaedalusVm MusicVm;
```
JaXt0r, line 23: I'd say it's inconsistent if we move MusicVm out of this class, but keep the other VMs cached in here.

Assets/Gothic-UnZENity-Core/Scripts/Globals/GameData.cs
```diff
@@ -16,13 +16,9 @@ namespace GUZ.Core.Globals
 {
     public static class GameData
     {
-        public static Vfs Vfs;
```
JaXt0r, line 19: Ok you moved this to ResourceLoader. Sounds reasonable. I think you aim to have this GameData class being removed in the future? What's you plan for v2 of this class (if any)?

Assets/Gothic-UnZENity-Core/Scripts/CoroutineManager.cs
JaXt0r, line 1: In C#, Interfaces start with I. Better: ICoroutineManager.
@see https://github.com/Gothic-UnZENity-Project/Gothic-UnZENity/blob/main/Docs/development/style-guide.md?plain=1#L88-L91

Assets/Gothic-UnZENity-Core/Scripts/CoroutineManager.cs
```diff
@@ -0,0 +1,11 @@
+using System.Collections;
+using UnityEngine;
+
+namespace GUZ.Core
+{
+	public interface CoroutineManager
```
JaXt0r: Do we really want to call it "Manager"? I didn't use this terminology for Interfaces before. A manager is more like "the one who's handling it" in my opinion. How about ICoroutine? Or IGUZCoroutine? (First best guess only ;) )

Assets/Gothic-UnZENity-Core/Scripts/Vm/VmInstanceManager.cs
```diff
@@ -0,0 +1,146 @@
...
+
+            return TryGetItemData(symbol.Name);
+        }
+
+        /// <summary>
+        /// Hint: Instances only need to be initialized once in ZenKit.
+        /// There are two ways of getting Item data. Via INSTANCE name or symbolIndex inside VM.
+        /// </summary>
+        [CanBeNull]
+        public static ItemInstance TryGetItemData(string key)
```
JaXt0r, line 37: I'm thinking if it makes sense for the using class, if we know _which_ VM we get the data from? What do you think? Currently it's not visible to the caller. I'm on the fence. I think we would need to split it, if we want to be more precise. Needed?
But I think you have only about 10 class types returning. Maybe we don't need to split this class into multiple ones for this topic.

Assets/Gothic-UnZENity-Core/Scripts/Vm/VmInstanceManager.cs
```diff
@@ -0,0 +1,146 @@
...
+using ZenKit.Daedalus;
+
+namespace GUZ.Core.Vm
+{
+    public static class VmInstanceManager
+    {
+        private static readonly Dictionary<string, ItemInstance> ItemDataCache = new();
+        private static readonly Dictionary<int, SvmInstance> SvmDataCache = new();
+        private static readonly Dictionary<string, SoundEffectInstance> SfxDataCache = new();
+        private static readonly Dictionary<string, ParticleEffectInstance> PfxDataCache = new();
```
JaXt0r, line 16: How about adding MusicTheme (from MusicVm) as well?

Assets/Gothic-UnZENity-Lab/Scripts/Handler/LabVobHandAttachPointsLabHandler.cs
```diff
@@ -80,9 +81,9 @@ private IEnumerator LoadVobOnClickDelayed()
 
         private GameObject CreateItem(string itemName)
         {
-            var itemPrefab = PrefabCache.TryGetObject(PrefabCache.PrefabType.VobItem);
-            var item = AssetCache.TryGetItemData(itemName);
-            var mrm = AssetCache.TryGetMrm(item.Visual);
+            var itemPrefab = ResourceLoader.TryGetPrefabObject(PrefabType.VobItem);
+            var item = VmInstanceManager.TryGetItemData(itemName);
+            var mrm = ResourceLoader.TryGetMultiResolutionMesh(item.Visual);
```
JaXt0r, line 86: I like this new way. One can clearly distinguish: Line 1+3 is from a Resource from Gothic. Line 2 is from VM data. 👍

Assets/Gothic-UnZENity-Core/Scripts/GameGlobals.cs
```diff
@@ -0,0 +1,41 @@
+using System;
+using GUZ.Core;
+using GUZ.Core.Manager;
+using GUZ.Core.Manager.Culling;
+using GUZ.Core.Manager.Settings;
+using GUZ.Core.World;
+
+namespace GUZ.Core
+{
+	public interface IGlobalDataProvider
```
JaXt0r: Please don't put multiple classes/interfaces into one file. I normally always do a lookup for file names when seeking classes etc. This _hidden_ interface made it hard to find.

Assets/Gothic-UnZENity-Core/Scripts/GameGlobals.cs
```diff
@@ -0,0 +1,41 @@
...
+		[Obsolete("Don't use globals.")] public static GameSettings Settings => Instance.Settings;
+		[Obsolete("Don't use globals.")] public static SkyManager Sky => Instance.Sky;
+		[Obsolete("Don't use globals.")] public static GameTime Time => Instance.Time;
+		[Obsolete("Don't use globals.")] public static RoutineManager Routines => Instance.Routines;
+		[Obsolete("Don't use globals.")] public static TextureManager Textures => Instance.Textures;
+		[Obsolete("Don't use globals.")] public static GUZSceneManager Scene => Instance.Scene;
+		[Obsolete("Don't use globals.")] public static FontManager Font => Instance.Font;
+		[Obsolete("Don't use globals.")] public static StationaryLightsManager Lights => Instance.Lights;
+		[Obsolete("Don't use globals.")] public static VobMeshCullingManager MeshCulling => Instance.MeshCulling;
+		[Obsolete("Don't use globals.")] public static VobSoundCullingManager SoundCulling => Instance.SoundCulling;
```
JaXt0r, line 24: I see. This is also the non-invasive way to keep the global lookup for now. What are your plans for v2? (We can also discuss on Discord, then you don't need to write it twice)

Assets/Gothic-UnZENity-Core/Scripts/GameGlobals.cs
JaXt0r: Can we also think about something which sounds more like "our" class? Whenever I had a class name, where I thought "ok, this could also be something from Unity itself", then I added the _GUZ_ prefix. Just, that you think about it at least.

Assets/Gothic-UnZENity-Lab/Scripts/LabBootstrapper.cs
JaXt0r, line 1: All meshes have black textures. It seems the MeshFactory.cs isn't used? Or at least the flag _useTextureArray_ is default (true), which won't render the textures on Vobs directly. But NPCs don't use TextureArray at all. 🤔

image

Assets/Gothic-UnZENity-Core/Scripts/Manager/Culling/VobMeshCullingManager.cs
```diff
@@ -39,29 +37,51 @@ private enum VobList
...
-        private readonly Dictionary<GameObject, Tuple<VobList, int>> pausedVobs = new();
-        private readonly Dictionary<GameObject, Rigidbody> pausedVobsToReenable = new();
-        private readonly Dictionary<GameObject, Coroutine> pausedVobsToReenableCoroutine = new();
+        private Dictionary<GameObject, Tuple<VobList, int>> pausedVobs = new();
+        private Dictionary<GameObject, Rigidbody> pausedVobsToReenable = new();
+        private Dictionary<GameObject, Coroutine> pausedVobsToReenableCoroutine = new();
+
+        private readonly CoroutineManager _coroutineManager;
+        private readonly bool _featureEnableCulling;
+        private readonly bool _featureDrawGizmos;
```
JaXt0r, line 46: So Manager now handle their _settings_ themselves. I like it. These properties makes it transparent, _which_ config values are used.

Assets/Gothic-UnZENity-Core/Scripts/Manager/Culling/VobMeshCullingManager.cs
```diff
@@ -14,7 +12,7 @@ namespace GUZ.Core.Manager.Culling
     /// With this set up, VOBs are handled by camera's view and culling behaviour, so that the StateChanged() event disables/enables VOB GameObjects.
     /// @see https://docs.unity3d.com/Manual/CullingGroupAPI.html
     /// </summary>
-    public class VobMeshCullingManager : SingletonBehaviour<VobMeshCullingManager>
+    public class VobMeshCullingManager
```
JaXt0r, line 15: Whenever we don't need more than 1 instance, we made classes static (Only reason was to show: Hey, I'm existing only once!).

Is there a use case why this class isn't static? We could also make this class static and call an _Init()_ function.

Assets/Gothic-UnZENity-Core/Scripts/Manager/Culling/VobMeshCullingManager.cs
```diff
@@ -39,29 +37,51 @@ private enum VobList
...
 
-            StartCoroutine(StopVobTrackingBasedOnVelocity());
+            _coroutineManager.StartCoroutine(StopVobTrackingBasedOnVelocity());
         }
 
         /// <summary>
         /// This method will only be called within EditorMode. It's tested to not being executed within Standalone mode.
         /// </summary>
-        private void OnDrawGizmos()
+        public void OnDrawGizmos()
```
JaXt0r, line 82: I didn't test, but I expect this won't work: OnDrawGizmos() is a Unity Editor function, which is called on all MonoBehaviours whenever we play in Editor mode.
You can also stick with _private_ as Unity will find this method anyways.

Assets/Gothic-UnZENity-Core/Scripts/Manager/Culling/VobSoundCullingManager.cs
```diff
@@ -1,25 +1,35 @@
...
-using GUZ.Core.Debugging;
 using GUZ.Core.Globals;
-using GUZ.Core.Util;
 using JetBrains.Annotations;
 using UnityEngine;
 
 namespace GUZ.Core.Manager.Culling
 {
-    public class VobSoundCullingManager : SingletonBehaviour<VobSoundCullingManager>
+    public class VobSoundCullingManager
```
JaXt0r, line 10: Same like VobMeshCullingManager.cs --> Why did you remove the SingletonBehaviour? We used it to tell Unity "this is a MonoBehaviour you'll get in the scene", but "only once". (I mark this topic as a general question as well.

Assets/Gothic-UnZENity-Core/Scripts/GlobalEventDispatcher.cs
```diff
@@ -2,13 +2,13 @@
...
 
-namespace GUZ.Core.Globals
+namespace GUZ.Core
 {
     /// <summary>
     /// Loading/Unloading order of scenes:
     /// https://github.com/Gothic-UnZENity-Project/Gothic-UnZENity/blob/main/Docs/development/diagrams/SceneLoading.drawio.png
     /// </summary>
-    public static class GUZEvents
+    public static class GlobalEventDispatcher
```
JaXt0r, line 11: Same like with the _IGlobalDataProvider_. Can we have something more meaningful which points towards: this is our game's class. Something like GUZGlobalEventDispatcher. (Do/Don't you like the GUZ prefix?)

Assets/Gothic-UnZENity-Core/Scripts/ResourceLoader.cs
```diff
@@ -0,0 +1,241 @@
...
+			var path = Directory.GetDirectories(root, "_work", new EnumerationOptions
+			{
+				MatchCasing = MatchCasing.CaseInsensitive,
+				RecurseSubdirectories = false
+			}).First();
+
+			return Path.GetFullPath(path, root);
+		}
+
+		private static List<string> FindDiskPaths(string root)
```
JaXt0r, line 240: This is great. Please add a comment with the default lookup/path structure for G1 as example. (Easier to understand "where" you look.)

Assets/Gothic-UnZENity-Core/Scripts/ResourceLoader.cs
```diff
@@ -0,0 +1,241 @@
...
+		[NotNull]
+		private static string GetPreparedKey([NotNull] string key)
+		{
+			var extension = Path.GetExtension(key);
+			var withoutExtension = extension == string.Empty ? key : key.Replace(extension, "");
+			return withoutExtension.ToLower();
+		}
+
+
+		private static string FindWorkPath(string root)
```
JaXt0r, line 223: This is great. Please add a comment with the default lookup/path structure for G1 as example. (Easier to understand "where" you look.)

Assets/Gothic-UnZENity-Core/Editor/Scripts/Builds/UnityBuilderAction.cs
```diff
@@ -37,7 +37,6 @@ static void PerformQuestBuild()
 		{
 			string target_path = TARGET_DIR + "/Quest/" + APP_NAME + ".apk";
 			SetQuestSettings();
-				FeatureFlagTool.SetFeatureFlags();
```
JaXt0r, line 40: This is ok. But we need to ensure we build with "Production" as GameConfig only. Or at least check what's there and highlight it in Build. Fine for now. But we need to think about it in the future.

I created a ticket to handle it afterwards: #56

Assets/Gothic-UnZENity-Core/Scripts/Creator/WorldCreator.cs
```diff
@@ -44,47 +41,45 @@ public static async Task CreateAsync(string worldName)
...
-            var zkMesh = zkWorld.Mesh.Cache();
-            var zkBspTree = zkWorld.BspTree.Cache();
+            var zkWorld = ResourceLoader.TryGetWorld(worldName);
             var zkWayNet = zkWorld.WayNet.Cache();
 
-            if (zkWorld.RootObjects.IsEmpty())
-                throw new ArgumentException($"World >{worldName}< couldn't be found.");
-            if (zkMesh.Polygons.IsEmpty())
-                throw new ArgumentException($"No mesh in world >{worldName}< found.");
+            if (zkWorld.RootObjectCount == 0) throw new ArgumentException($"World >{worldName}< couldn't be found.");
```
JaXt0r: Please don't put single IF-commands into the same line. We wanted to have them with braces on a separate line.

Assets/Gothic-UnZENity-Core/Scripts/Creator/WorldCreator.cs
```diff
@@ -93,10 +88,11 @@ private static void LoadWorld(string worldName)
                 WayNet = (CachedWayNet)zkWayNet
             };
 
-            GameData.World = world;
+            return world;
         }
 
-        private static async Task<List<WorldData.SubMeshData>> BuildBspTree(IMesh zkMesh, IBspTree zkBspTree)
+        private static async Task<List<WorldData.SubMeshData>> BuildBspTree(IMesh zkMesh, IBspTree zkBspTree,
```
JaXt0r, line 100: About _lightingEnabled_: For the other Managers you used the GameConfig to set it via Init() and stored it's values inside bool _feature*_ Fields.

Assets/Gothic-UnZENity-Core/Scripts/GameConfiguration.cs
```diff
@@ -0,0 +1,99 @@
...
+		[SerializeField] public bool enableNpcRoutines = false;
+
+		[SerializeField] public string spawnAtWaypoint = string.Empty;
+
+		[SerializeField] public bool enableWorldObjects = true;
+		[SerializeField] public bool enableWorldMesh = true;
+		[SerializeField] public bool enableBarrierVisual = true;
+		
+		[InspectorName("Experimental: Enable Decal Visuals")]
+		[SerializeField] public bool enableDecalVisuals = true;
```
JaXt0r: Should be: _false_ by default / for Production builds.
Hint: If you don't want to reflect "production" in the class itself, then let's stick with C#'s default values in general and we always _clone_ the Production.asset one whenever we need a specific instance.
(I won't comment more elements on this for now. I think the idea with C# default values would help. Let's discuss.)

Assets/Gothic-UnZENity-Core/Scripts/GameManager.cs
```diff
@@ -0,0 +1,185 @@
...
+		private bool _isInitialised = false;
+
+		public GameSettings Settings => _settings;
+		public GameConfiguration Config => config;
+		public SkyManager Sky => _skyVisualManager;
+		public GameTime Time => _gameTimeManager;
+		public RoutineManager Routines => _npcRoutineManager;
+		public TextureManager Textures => _textureManager;
+		public GUZSceneManager Scene => _gameSceneManager;
+		public FontManager Font => _fontManager;
```
JaXt0r: That's funny. I didn't know, that we can define a field _private_ and as the appropriate Interface Field is marked as _public read_, it will override. It's kind of intransparent tbh. Could you please clone the public/private field values from the interface to these elements, please.

Assets/Gothic-UnZENity-Core/Scripts/GameManager.cs
```diff
@@ -0,0 +1,185 @@
...
+using GUZ.Core.Util;
+using UnityEngine;
+using UnityEngine.SceneManagement;
+
+namespace GUZ.Core
+{
+	[RequireComponent(typeof(TextureManager), typeof(FontManager))]
+	public class GameManager : MonoBehaviour, CoroutineManager, IGlobalDataProvider
+	{
+		public GameConfiguration config;
```
JaXt0r: Is the intendation broken by +4 spaces on the following lines?

Assets/Gothic-UnZENity-Core/Scripts/GameManager.cs
```diff
@@ -0,0 +1,185 @@
...
+			if (!_settings.CheckIfGothic1InstallationExists())
+			{
+				invalidInstallationPathMessage.SetActive(true);
+				return;
+			}
+
+			// Otherwise, continue loading Gothic.
+            ResourceLoader.Init(_settings.GothicIPath);
+	
+            _gameMusicManager.Init();
```
JaXt0r: Intendation issue?

Assets/Gothic-UnZENity-Core/Scripts/GameManager.cs
```diff
@@ -0,0 +1,185 @@
...
+			_meshCullingManager = new VobMeshCullingManager(config, this);
+			_soundCullingManager = new VobSoundCullingManager(config);
+			_barrierManager = new BarrierManager(config);
+			_stationaryLightsManager = new StationaryLightsManager();
+			_xrSimulatorManager = new XRDeviceSimulatorManager(config);
+			_gameTimeManager = new GameTime(config, this);
+			_skyVisualManager = new SkyManager(config, _gameTimeManager, _settings);
+			_gameMusicManager = new MusicManager(config);
+			_gameSceneManager = new GUZSceneManager(config, _gameLoadingManager, xrInteractionManager);
+			_npcRoutineManager = new RoutineManager(config);
```
JaXt0r: That's the _new()_ behaviour of Managers I also asked in general. What's the benefit of it?

Assets/Gothic-UnZENity-Core/Scripts/GameManager.cs
```diff
@@ -0,0 +1,185 @@
...
+				return;
+			}
+
+			_isInitialised = true;
+			Load();
+		}
+
+		private void FixedUpdate()
+		{
+			_barrierManager.FixedUpdate();
```
JaXt0r: Barrier manager should take care of it. Do you really want to centralize these Unity calls? I know your point about having nearly 0 managers as a Component, but I don't think this will scale properly. Let's discuss.

Assets/Gothic-UnZENity-Core/Scripts/Manager/GUZBootstrapper.cs
```diff
@@ -6,127 +6,47 @@
...
-#pragma warning disable CS4014 // It's intended, that this async call is not awaited.
-                GUZSceneManager.I.LoadStartupScenes();
-#pragma warning restore CS4014
-            }
-            else
-            {
-                // Show the startup config message.
-                invalidInstallationDirMessage.SetActive(true);
-            }
-        }
```
JaXt0r, line 64: Thanks a lot for removing this one. I actually implemented this and hated it at the same time. 🤣

Assets/Gothic-UnZENity-Core/Scripts/Resource.cs
```diff
@@ -0,0 +1,41 @@
+using System;
+using System.Collections.Generic;
+using JetBrains.Annotations;
+using UnityEngine;
+using ZenKit;
+
+namespace GUZ.Core
+{
+	public class Resource<T>
```
JaXt0r, line 21: I like the way you handled this Typed loading properly.
Please add a class comment onto this class and the ResourceLoader class to explain how to use. In the end we normally don't need to understand the details of this magic, but with a comment on both classes, you help people on "howto use it.

JaXt0r (commented):
Hi @lmichaelis . Nicely done. I added my remarks at the specific code parts. Feel free to comment, use, or decline at your own judgement.

Tested:
* Lab

Open to test:
* Normal game



A few general questions about the source code:
1. You removed the SingletonBehaviour<?> from all/most of the *Manager.cs classes. Whenever we don't need more than 1 instance, we made classes static (Only reason was to show: Hey, I'm existing only once!). Is there a use case why this class isn't static? We could also make this class static and call an Init() function.
2. Some classes like _GlobalEventDispatcher_ or _IGlobalDataProvider_ don't sound like our classes. I suggest we think about a common prefix for class names, which sound like Unity ones. Goal: To distinguish easily between _our_ logic and _Unity's_. I'm happy to change my mind, but want to discuss it with you first.
3. GameManager is handling all the Unity events like Update(), LateUpdate(), FixedUpdate(). My first idea was: The managers should take care of it on their own. Can you explore the benefits and drawbacks a little bit better, please. I'm on the fence right now.

lmichaelis:
> 1. You removed the SingletonBehaviour<?> from all/most of the *Manager.cs classes. Whenever we don't need more than 1 instance, we made classes static (Only reason was to show: Hey, I'm existing only once!). Is there a use case why this class isn't static? We could also make this class static and call an Init() function.

The idea here is to make the dependencies explicit. So, if a manager requires another manager to function correctly, we need to make sure that the other manager is properly initialized before its used. Passing them explicitly (-> dependency injection) helps make that dependency obvious to anyone inspecting the code.

> 2. Some classes like _GlobalEventDispatcher_ or _IGlobalDataProvider_ don't sound like our classes. I suggest we think about a common prefix for class names, which sound like Unity ones. Goal: To distinguish easily between _our_ logic and _Unity's_. I'm happy to change my mind, but want to discuss it with you first.

Hm maybe. I'd suggest we discuss a naming scheme internally and then add that to the style guide. Then we can rename all of the existing classes (since we'll probably need to rename others too).

> 3. GameManager is handling all the Unity events like Update(), LateUpdate(), FixedUpdate(). My first idea was: The managers should take care of it on their own. Can you explore the benefits and drawbacks a little bit better, please. I'm on the fence right now.

With this setup, the managers cannot take care of this because they're not `MonoBehaviour`s. This means, that this setup is required. It is also nice to be able to control the order in which the `Update` (and so on) functions are called directly and make it visible immediately which managers need to be updated and which ones don't.

JaXt0r:
Hi @lmichaelis I thought about it and fully agree for the managers:
1. Let's stick with your DI idea. I see your point in using the proper init order and centralized initialization of these singleton managers. It can also have benefits in the future if we want to overwrite manager classes for specific behaviour (e.g. G1/G2, XRIT/HVR, ...).
2. It might also be just me who's unconfident with this. In the end, I will get used to it once settled. The more I think, the more I dislike prefixes as they're redundant and make class/file names even longer. Feel free to suggest some naming schema or we just go on with your options for now (Especially as the new classes are in the root folder of Core/Scripts and easily seeable.)
3. Ok. Understood. Let's stick with this central management of Managers (what a phrase 🤣). We need to act decentralized for Components (e.g. colliders on NPCs) but this is another topic and has nothing to do with the managers' centralized handling itself.

Please add a few information on the docs about the Manager handling and where they're initialized, then I'm fine about the three topics above.

Assets/Gothic-UnZENity-Core/Scenes/Bootstrap.unity
JaXt0r, line 1: Texturing of Vobs is wrong. The image frame in Gomez' throne room should show the hero fighting with a Harpyie. Last time we had this bug, I might remember, that we loaded the VFS files in the wrong order (Oldest vs Newest).

image

JaXt0r:
I tested the main game including MainMenu, NPCs, and world switching. Except from the two texture issues (Textures of NPCs+Vobs in Lab scene + image frame in Gomez' throne room), everything works fine! Please go on.

Assets/Gothic-UnZENity-Core/Scripts/ResourceLoader.cs
lmichaelis: Indeed it should not be cached because when we load a save-game for example, the VM needs to be reloaded entirely. We should manage VMs somewhere else since they need special setup anyways.

lmichaelis: Maybe in the future but then somewhere else. The resource loader is just for loading resources and caching immutable ones.

Assets/Gothic-UnZENity-Core/Scripts/Globals/GameData.cs
lmichaelis: It is but I will remedy this in another PR. The whole codebase is in an inconsistent state right now.

Assets/Gothic-UnZENity-Core/Scripts/Globals/GameData.cs
lmichaelis: Yes, correct. My plan is to remove `GameData` entirely and replace it with a `World` component. We'll prefab that and then remove the need for a separate scene for each world.

Assets/Gothic-UnZENity-Core/Scripts/Vm/VmInstanceManager.cs
lmichaelis: Yes, this is a temporary solution.

Assets/Gothic-UnZENity-Core/Scripts/Vm/VmInstanceManager.cs
lmichaelis: I don't want to add that here since I intend on removing this entirely anyways. It's just a stopgap solution.

Assets/Gothic-UnZENity-Core/Scripts/Manager/Culling/VobMeshCullingManager.cs
lmichaelis: Yes but this class is not a mono behaviour any more.

Assets/Gothic-UnZENity-Core/Scripts/Creator/WorldCreator.cs
lmichaelis: Not here because I prefer not to have internal static state here.

Assets/Gothic-UnZENity-Core/Scripts/Globals/GameData.cs
JaXt0r: Unfortunately we need these different world scenes for culling. Unity hard wires scenes and their mesh culling logic based on a folder with the same name as the scene.

Assets/Gothic-UnZENity-Core/Scripts/ResourceLoader.cs
JaXt0r: Please add a warning, that this is a compute intensive call and point to where you cashed them as of now. Sure, you'll change, but at least highlight it for now.

Assets/Gothic-UnZENity-Core/Scripts/Manager/Culling/VobMeshCullingManager.cs
JaXt0r: Yes. Understood in a discussion with you yesterday. 🫡

Assets/Gothic-UnZENity-Core/Scripts/CoroutineManager.cs
lmichaelis: Not sure, we'll revisit this later, as I said before, when we'll set up a guide for naming things

JaXt0r: Check. But at least add the >I< prefix, please. Then we see, it's a "manager" but the interface for it. Then I'm happy with it.

Assets/Gothic-UnZENity-Lab/Scripts/Handler/LabMusicHandler.cs
lmichaelis: It works for now, so we'll leave it I think. Later we can create a `MusicManager` instance for the lab.

Assets/Gothic-UnZENity-Lab/Scripts/LabBootstrapper.cs
lmichaelis: @JaXt0r This is the last issue standing. I don't know what the problem is here since I didn't change any Lab logic. Also, you might just have to reload the scene because it works for me:

image

JaXt0r:
Looks great. I will check with the lab issue. I also think it's related to my opened Unity.

PR #43, opened 2024-06-17
Bugfix - Check if Daedalus Symbol has return type

### LINKED ISSUE #36: Daedalus VM crashes with OC NPCs with Stackoverflow exception
Currently NPC routines fill up the Daedalus stack and cause a StackOverflowException after a few minutes. We should print the stack after a few seconds to see, which method fills it up. This might be related to an external function which adds parameters which aren't fetched on our side.

JucanAndreiDaniel:
Fixes a crash in the AiHandler.


JaXt0r (approved):

JaXt0r:
Is your PR is really fixing the crash (https://github.com/Gothic-UnZENity-Project/Gothic-UnZENity/issues/36) But beware: If got a StackOverflowException. Not a NonExistantReturnValueException.
Did you check if it is solved by running the game with all (OC) NPCs for about 5+ minutes?

PR #53, opened 2024-06-21
Feature/npc spawn on top of each other

### LINKED ISSUE #15: NPCs spawn on top of each other
Image

We need to spawn them "next to each other" like in G1

If we spawn NPCs, we spawn them all at the same frame. It means Unity's Physics logic can't handle a Physics.CheckCapsule() at that frame (Physics run at FixedUpdate() which is _after_ the GO spawn time). We should therefore check for two logics:
1. When we spawn an NPC, we need to look up if there's an NPC already spawned at that spot in this frame. If so, then circle around it and find a new spot via offset. And store this NPCs location in a temporary list until the next cycle of Fixed update().
2. Physics.CheckCapsule(). If space is blocked. If so, circle around and store the value in the temporary list for this Fixed update() cycle.

Alternatively:
* Set Rigid body position in the spawning frame and see if it will be recognized by Physics directly. (https://forum.unity.com/threads/how-to-properly-set-initial-position-of-rigidbody.1464332/)

Hint: be careful with space below the NPC. Choose capsules lowest y value carefully.

JaXt0r:
> Please test now, but merge only after https://github.com/Gothic-UnZENity-Project/Gothic-UnZENity/pull/39

# To test:
1. Start game with all NPCs active
2. Stop immediately after they are spawned
3. Check the locations I mentioned in screenshots if the NPCs spawn _next to each other_ instead of _onto each other_

OC: Mages
image

Bloodwyn and Ball Taran:
image

Hint:
* If you compare to G1 closely, you will see, that the rotation angle of where to look for free spots is different. But tbh: who cares. 🤷 
image


lmichaelis (approved):
Works as described. Only one thing I noticed: Not all NPCs seem to be spawning correctly:
 
image

JaXt0r:
> Works as described. Only one thing I noticed: Not all NPCs seem to be spawning correctly:
> 
> image

Good finding. Thank. It means the "circle search" isn't fully working so far. I will update and see, that no NPC will spawn at 0,0,0!

JaXt0r:
I fixed it (+added an error message if no spot was found) and will merge now. If anything's fishy, I can easily fix later.

PR #61, opened 2024-06-29
chore: reformat code according the style guide

lmichaelis:


JaXt0r:
Hi Luis. Can you briefly summarize which types of changes you did? E.g. intendation, element ordering, casing, file renamings, ...
This makes it easier for me to check.

As a bonus, it would be great if you can briefly describe - e.g. with images - what you clicked in Rider to execute the changes.

lmichaelis:
I used autoformat for most of it, which added missing braces, sorted imports, removed redundant code (such as some unnecessary casts and imports) and made sure everything is layed out properly. I then did a separate step of renaming which mostly affects fields, changing them to the proper format (e.g. PascalCase for public and protected fields and underscore-prefixed camelCase for private fields). This did **not** change any of the field from public to private, that'd be a manual action.

To format in Rider: Press `Ctrl+Shift+Alt+L`, then select a scope, then either select a pre-defined profile or click the small edit button in the "Profile" column to configure the formatting (that's what I did):

image

In the popup click "Add" to create a new formatting profile, then enter a name and configure it to the left:

image

The important bits here are: Disable everything under "Rearrange Code",  under "Optimize Imports", make sure not to select "Embrace 'using' directives in region" (the other options are good), and under redundancies, only select "Remove code redundancies" (which I'd recommend). I'd also recommend enabling all options under "Apply syntax styles".

image

Then, save your configuration, select it in the initial popup, and click "Run". Done.

JaXt0r (approved):
I checked some sample classes. Looks good. Thanks.

I have just one request: The documentation for Rider setup to "auto apply" our wanted changes only (imports, unneccessary spaces, braces around IFs, ...) are very helpful. Please put them into our style-guide.md to make it easier for us to apply them regularly.

Issue #23, opened 2024-06-10
HVR: Door opening - Update physics


JaXt0r:
Doors can't be opened as of now. Let's change this.

## TODOs
- [x] Doors are ocMobInteractable, separate them in VobCreator via name, so that they will get their own prefab -> Added with #57 
- [x] Inside prefab, implement the sub-GOs via regex-wildcards and add the HVR components -> Added with #57 
- [x] The doors are by default physics based. We need to find a way to ignore walls as it might be the door will stuck inside world (e.g. door knob will collide with wall)
- [x] (Bonus): Animation of door opening has a sound. Check in animation data which rotation % it will play and add it as event to the door.

![Image](https://github.com/Gothic-UnZENity-Project/Gothic-UnZENity/assets/120568393/0ae1679a-9a09-4450-806b-7c144d01f593)

JaXt0r:
Partially done via: #81

Just sound is missing.

Issue #19, opened 2024-06-10
HVR: Chest opening


JaXt0r:
Chests contain a separate mesh for their lid. We can leverage this to create open and close animation properly.

## TODOs
- [x] Update prefab of ocMobContainer and add a sub-element "BIP01 CHEST_.*_0"."BIP01 CHEST_.*_1" (Prefab GOs are replaced via regex with loaded ZenKit data).
- [x] Add the neccessary HVR components to the prefab GO -> Added with https://github.com/Gothic-UnZENity-Project/Gothic-UnZENity/issues/57
- [x] Adjust parameters to enable only partial opening of the lid. (I assume the same value can be applied for all chests as they're very similar) -> Added with https://github.com/Gothic-UnZENity-Project/Gothic-UnZENity/issues/57
- [x] The chests are by default physics based. We need to find a way to ignore walls as it might be the chest lid will stuck inside world (e.g. lid will collide with a wall when opened)
- [x]  (Bonus): Animation of chest opening has a sound. Check in animation data which rotation % it will play and add it as event to the chest.
- [x] (Bonus): Let the lid close automatically when we move away from it. (G1 itself will never leave chests open.)

![Image](https://github.com/Gothic-UnZENity-Project/Gothic-UnZENity/assets/120568393/ff3b7877-e029-4f26-bfc5-c9e00078290d)

JaXt0r:
Only open topic: The lid should stay open when we're very close. Otherwise we always need to grab with one hand while looting.

JaXt0r:
Solved via: #81

PR #66, opened 2024-07-05
Highlight objects pointed at + show its name

### LINKED ISSUE #57: Show Vob/NPC name and highlight it when pointing towards
In G1 classic, an object is highlighted brighter with ist's name on top of it, once marked/highlighted/pointed towards. Good thing is, that HVR recognizes "pointing" towards an object out of the box (we can see it with the default remote grabbing feature of items). We now need to extend this feature.

- [x] Understand and reuse HVR remote grabbing feature (as it already calculated the "pointing" of our hands towards an object and should be reused/extended if possible.)
- [x] Brighten up the highlighted object (shader? etc.? Please ask on Discord for ideas.)
- [x] Print name of object (Axe, Bloodwyn, ...) on top of it. (Either put it on the actual prefab or calculate "top" based on object collider's bounding box)
- [x] Implement logic for
    - [x] oCItem
    - [x] oCMobLadder
    - [x] oCMobFire
    - [x] oCMobContainer - currently trying to act as physical objects and clip through world!
    - [x] oCMobDoor - currently trying to act as physical objects and clip through world!
    - [x] oCMobBed
    - [x] oCMobWheel
    - [x] oCMobSwitch
    - [x] oCMobInter
- [x] Optimize label positions to work properly for Ladders (very long) and various small and big items.
- [x]  Bonus: Print also amount of item next to its name. e.g. Beer (3x)

JaXt0r:
Like G1, marked objects are now brightened up and their name is shown.

To test:
* General expected behaviour: Objects should be highlighted most of the time when pointed towards.
* Spawn Lab with HVR (+/- Simulator) and test the oCMob* elements left of ladder demo.
* Spawn world and try if names and highlights are shown on objects as well
    * Object names should rotate along your camera movement
* GameConfiguration has two new fields: _BrightenUpHoveredVOBs_ + _ShowNamesOnHoveredVOBs_ -> Please disable them and see if it works as you expect it.

Boundary:
* Inside lab, VOBs have no name
* Some objects also have no name inside G1 (e.g. lights on the street). It's intended behavior by G1.
* HVR "Hover" feature sometimes doesn't recognize, that we point to an object. We need to investigate it later if it disturbs.
* It might be, that not all VOB names are perfectly visible from all angles. Please judge for yourself if it's a blocker or ok for now.
* Separate tickets:
    * Doors and Chests have the HVRPhysicalDoor component and rotate already. But it's still physics based and will stop when a world collision (e.g. a wall) is detected. I will mention this in the Door/Chest tickets to solve there. (#19 + #23)
    * Every object has HVRGrabbable as component to use its Hover events. Even if we won't "grab" an object (e.g. a lever). I will create a separate ticket to think about a better logic. (#67)
    * Some objects (e.g. water pipe) have incorrect Grabbable logic (prefab-ed on wrong spot). We will create a new ticket for better lookup during mesh creation in a separate ticket. (#68)


lmichaelis (approved):
Tested the Lab and Main; highlighting works and names show up correctly. Config works as expected. Skimming the code, it looks okay too.

Only one problem remains: VOb culling does not work; VObs with custom prefabs will show up as purple.

JaXt0r:
Thanks for testing. I created a follow up issue for the textures: https://github.com/Gothic-UnZENity-Project/Gothic-UnZENity/issues/72

PR #74, opened 2024-07-08
Purple Material fix for OC'd VOBs with HVR 

### LINKED ISSUE #72: Occlusion Culling for VOBs removes Meshes when HVR Player is used
It's only reproducible on normal game start (e.g. world.zen):
* with Culling activated
* and HVR player activated (tested with Simulator)

Image

It affects the TextureArray elements (e.g. fire on start area) and also single meshes (e.g. torch on the ground).

JaXt0r:
To test:
* Load game with VOBs and Occlusion Culling activated.
* Rotate around
* Check if VOBs stick with their normal texture (ok) or get back their purple one (not ok)


lmichaelis (approved):
Works!

PR #89, opened 2024-08-20
Feature/80 gothic modding workshop borobongo turn to npc

BoroBongo:
This feature adds a script to the VRPlayer GameObject that moves the VRPlayer Root Transform to the Camera's transform. Because of this the NPC (Orry) turns in accordance to us and not where the VRPlayer got spawned in the world. 

To get this working you'll need to add the `MoveTranform.cs` script to the VRPlayer GameObject - everything else should happen automatically once the game fully initializes


JaXt0r:
@borobongo thanks for your PR.

I tested it. Leveraging Camera.transform as Head for the NPC is a good idea to check if there's a line of sight between an NPC and the hero. Unfortunately there's a bug as Physics.Linecast checks if any collider is between position1 and position2. But the camera has no collider. Only the shoulders of the hero. ;-)

You can see the shoulders as boxes in my screenshot and the camera position as sphere.
So I think it makes sense to
1. add a Collider onto the camera (non collisioning, just to fetch the linecast data
2. do a linecast check with HitInfo to see if the correct collider was called.

image

JaXt0r (changes requested):
.

Assets/Gothic-UnZENity-HVR/Resources/HVR/Prefabs/MoveTransform.cs
JaXt0r, line 1: I think there is no need for this class any more. I did the fix for NpcHelper without activating this class.

Assets/Gothic-UnZENity-Core/Scripts/Manager/NpcHelper.cs
```diff
@@ -32,6 +32,7 @@ public static void CacheHero(GameObject playerGo)
             // Set data for NPC.
             var npcInstance = (NpcInstance)GameData.GothicVm.GlobalHero;
             playerProperties.NpcInstance = npcInstance;
+            playerProperties.Head = Camera.main.transform;
```
JaXt0r, line 35: Thanks for adding this. I checked it and this transform seems to be exactly at the height of the player. Finally. :-)

JaXt0r (approved):
Rebased parent branch and tested it. Works properly. Can be merged.

PR #90, opened 2024-08-25
Fix/turn to npc better anim

BoroBongo:
- Tweaked the speed of head turn;
- Added idle animation to after TurnToNPC is finished


JaXt0r (approved):

JaXt0r (approved):
I fixed the Diego bug. Now the head rotation and NPC rotation looks pretty decent. Ready to be merged. 👍

PR #92, opened 2024-08-26
Add transparent single mesh shader & fix SRP compat for opaque shader

ThomasMountainborn:
All properties need to be in the UnityPerMaterial cbuffer, otherwise you break SRP batching compatibility. You can see this in the inspector when selecting the shader in the editor.

When switching the mesh to the transparent shader, make sure the render queue of the material instance is also set to transparent, and revert it to geometry when switching the shader back to opaque.


JaXt0r (approved):
Transparency works well. The open topic if we should leverage another Lit-World-Dynamic shader can be changed later.

PR #98, opened 2024-08-27
update(GUZHVRPlayerInput): Jump with right "A" button implemented.

JaXt0r:
## To Test
* Hit "A" button on your right controller and the player will jump.

PS: Smalles PR in our project's history. 🤣


JucanAndreiDaniel (approved):
Works fine on device

PR #112, opened 2024-09-05
add(VRVobDoor): Added sounds for opening/closing door.

### LINKED ISSUE #23: HVR: Door opening - Update physics

JaXt0r:


JaXt0r:
I double checked the code and tested it.

PR #115, opened 2024-09-06
Feature/some pipeline testings

JaXt0r:


JaXt0r:
Pipeline now runs on every PR creat+update+reopen run and checks for HVR setup. NICE!

PR #119, opened 2024-09-11
Feature/npc drax walk idle dialog

JaXt0r:
## Fixes:
* Idle animation for walking NPC is now S_*WALK (before that: S_*WALKL did a walk instead of a breathing)

## To test
- [ ] Spawn Ratford (818) and Drax (819). They should talk now. Once they did talking and wait for 3-6s, they will breathe only (Before that: They walked into each other)


JaXt0r:
I need to merge this into main, as I will rewrite scene loading logic these days. (#AfraidOfMergeConflicts)
I double checked the changes here.

I also disabled perceptions for now and added a FIXME comment.

PR #129, opened 2024-09-15
Optimized resource usage

ThomasMountainborn:
From going out of memory when loading Gothic 2 on Quest 2, to having almost 2 GB of free memory left.


JaXt0r (approved):
🚀

JaXt0r:
I will merge it now as we need these improvement on the new G2 branch.

PR #134, opened 2024-09-18
fix(AbstractMeshBuilder): MeshCache is now handling multiple meshes p…

JaXt0r:
…er object (e.g. NPC armor has multiple base meshes and attachment meshes)


JaXt0r:
Checked the changes on whole G1 world scene loading.
NPCs now have proper armors again.
Code changes affect the needed changes only.

PR #135, opened 2024-09-19
Feature/gothic 2 initial support

### LINKED ISSUE #97: Architectural support for Gothic 2
As this is a foundational ticket, the goal should be:
* Boot the game and have the hero being spawned at start position
* Nothing more. The game might be broken but it shouldn't crash for the first seconds when loaded. ;-)

From a foundational development perspective I see the following changes we need to do to our core logic:
- [ ] User handling
    - [x] GameConfiguration.json needs to support a JSON-property for Gothic2 installation directory as well
    - [x] When we boot the game for the first time on a MVR device, we need to create an empty G2 folder as well (for convenience reasons)
    - [ ] User needs to have an additional menu before booting the game which asks them if they want to play Gothic1 or Gothic2 (This defines which Context will be selected and the folder of the game to choose from)
- [ ] Script logic distinguishment between G1 and G2
    - [x] Create two new modules: _UnZENity-Gothic1_ and _UnZENity-Gothic2_
    - [x] We need to enhance our GUZContext.cs logic to add Adapter for G1 and G2 as they have different logic in some areas. (Hint: Even if it seems overkill, we will love it, at least when we also add Mod support for Jinja stuff etc. Then we have another module to handle and select via context)
    - [x] e.g. The ini file of G2 might contain different entries then G1's INI file. Therefore it might be, that we need to have 2 implementations of parsing logic.
- [ ] Scenes + Occlusion Culling (OC)
    - [ ] OC tool for maps needs to be fixed (I think it's broken for some time)
    - [x] Map scenes with OC data needs to be put into a separate folder: Scenes/Gothic2/...

PS: This issue should be moved to another GitHub project: _4 - G2 support_ (Can be created if we have more than this ticket in the future).

JaXt0r:
Changes so far:
* GameSettings.json now supports Gothic1 and Gothic2 paths
* GameConfiguration now supports flag for Context.GameMode=G1|G2
* Module for Gothic1 and Gothic2 are added (mostly skeletons)
* Some (I remember 4) small fixes to load Gothic world added

On top of that: No further Gothic2 changes are done so far. It's just: The world loads.


JaXt0r:
I need to merge this into main, as I will rewrite scene loading logic these days. (#AfraidOfMergeConflicts)
I double checked the changes here.

PR #137, opened 2024-09-22
Feature/117 refactor scene handling

### LINKED ISSUE #136: Refactor: Scene loading logic - 2024
## Description
We currently leverage the following scenes: Bootstrap, MainMenu, Loading, General, and world scenes.
As our logic _before_ the main game loop grows (G1/G2 select, PreCaching of meshes), we need an adjusted scene handling.

My current workflow handling it:
Image

Basic changes are:
* We keep the Player scene (VR Player) active and alive all the time. (Because in VR, we even want players to rotate their heads in each scene, even loading; currently we need to load them manually in each scene.)
* We have two new initial scenes: GameModeSelect (Choose between G1 and G2) and PreCaching (the new world caching logic)
* Both scenes are optional. It means if there's only one Gothic installation set and the cache exists, the game will automatically go into MainMenu scene

Justification:
* I hope that loading of Player into the game once is made easier as we currently load and unload it based on main scene
* We need to handle two new scenes which have different UIs (GameMode + PreCaching). I therefore think of using the scenes which load and unload asynchronously
* Bootstrap scene (and its *Manager.cs) will have clear APIs to load scenes and handle next scene loading

JaXt0r:
## Description
Scene handling moved from GUZSceneManager towards separate *SceneManager classes which call LoadScene and UnloadScene directly. This lowered complexity.

## To test
* Set some settings inside GameConfiguration (with/without main menu, spawn at specific VOB)
* Check if the game boots normally and if you're spawned at the right spot with VR Player

## Technical details
Diagram about changes can be found here: https://github.com/Gothic-UnZENity-Project/Gothic-UnZENity/wiki/Developer-Hints#scene-loading


JaXt0r:
@BoroBongo tested it twice. I fixed the remaining topics. Also Teleportation via Hands is working properly. The other fixes and findings are named in the commits.

PR #140, opened 2024-09-24
add(GameConfiguration.PreselectWorldToSpawn): Define world to spawn player into.

JaXt0r:
Please check if altering this dropdown is providing the expected outcome.

image


JaXt0r:
Tested and works. I need to merge it now, as I rewrite world loading in PreCacheScene branch already.

PR #141, opened 2024-09-24
Bugfix/use vr device in editor mode

JaXt0r:
## Done
* Updated OpenXR to LATEST
* Removed XRIT (Package manager + demo assets) fully.
* Re-created OpenXR assets from scratch (/Assets/XR/*)

## To Test
* Simply start the game once (Production settings are fine) --> Is it rendering inside your device?


JaXt0r:
I tested it and it works. Need to merge it now, as we have too many differences in various PRs these days.

PR #144, opened 2024-09-25
Feature/SkinnedMeshRenderer - use TextureArray as well

JaXt0r:
Vobs with SkinnedMeshRenderer (e.g. WaterPipe) are now using TextureArray as well.

image


JaXt0r:
As discussed with Thomas: This change only affects static VOBs which are loaded during world loading time. No change for NPCs or items loaded at a later state. (Currently the change only affects water pipe in G1)

PR #146, opened 2024-10-01
Feature/117 pre cache meshes

### LINKED ISSUE #117: Cache world mesh + static VOB mesh to be calculated only once
## Background
Loading world.zen currently takes about two minutes (on a Quest2). Most of the time is consumed by calculating world mesh chunks and assigning the mesh (etc.) to the proper game objects.

## Solution
We can store the created objects and its components in cache files. Either via plain YML/JSON/binary files or to something sophisticated like glT where we can Export/Import whole GOs+Components at runtime.

## Tasks
- [ ] Build cache store mechanism
  - [x] ~Alter GameObject creation to not create Unity GameObjects, but store the neccessary data into lists and store them later (do not render data - save performance and unneccesary memory usage)~ --> **Out-of-Scope:** This would require a whole rewrite of VobManager and WorldCreator logic. Skipping for now.
  - [x] Add header information with version flag. i.e. if we change something on the data to cache, this will be checked and cache will be recreated automatically
  - [x] Explore options on how to save+retrieve async without frame impediments (e.g. save into multiple files, read a chunk of bytes per frame only, check if non-main thread is sufficient to not block main game loop) --> **FIX:** Task.Run() is saving and loading data in another thread. Fully async and non-blocking.
  - [x] Store duplicate meshes for VOBs as one Mesh only and reference it during loading
  - [x] Check if Binary read+write is better or JSON based (file storage size + file read times) --> **FIX:** We store via Unity's JsonSerialize and GZIP the data before storing. G1 all Worlds+static VOBs: ~70MB (similar to glTF solution size tested before)
  - [x] Explore compression of cache files. Measure how it impacts save+load performance and storage needs. --> **FIX:** GZIP applied to FileStream when saved. Shirnks files by factor 10. (whole G1 caching without it ~600MB with ~60MB)
  - [ ] Support SkinnedMeshRenderer for VOB _Waterpipe_ in G1. (Add bones and bone weights.)
- [ ] Build cache retrieve mechanism
  - [x] Retrieve data in a non-blocking way (no frame drops on player movements itself) --> **FIX:** Tas.Run() as well
  - [ ] Vobs: _VobCacheManager_ loads them without Prefab usage and no additional Components (i.e. only _MeshFilter_+_MeshRenderer_). We need to loop through them again and apply Prefab elements (most likely: Clone components from cached data into new objects)
  - [ ] Vobs: Load additional Vobs which are not from cache (e.g. _ZoneMusic_) and place them inside same Hierarchy. To be handled by _VobManager_
- [ ] Misc
  - [ ] Create proper Loading Manager to highlight status of loadings (ProgressBar fix)

## Links
* BinaryFormatter vs JSON serialization: https://medium.com/nerd-for-tech/how-to-implement-serialization-and-deserialization-in-c-f783e7fc0901

JaXt0r:
## Description
TBD

## Out-of-Scope


JaXt0r:
Some updates on the PreCaching logic:
* We save 2 seconds of loading time compared to previous logic of loading. 😕
* I can still optimize Phase 2 a little, but I don't expect huge frame benefits

This is the data for now:

**old logic without cache** 
* 26s/27s - full world + VObs loading - (but unoptimized for fps - aka there are hickups and if optimized it would take about 30-35 I assume)

**new logic with cache**
* 24/25 - full world + VOB loading
* ->  7s - Phase 1 - Cache loading from file and apply Meshes to scene
* -> 12s - Phase 2 - Apply prefab to cached VOB meshes - Prefabs applied to 11.286 VOBs
* ->  6s - Phase 3 - Create dynamic/non-meshed VOBs - created new VOBs 3.267

**Some points against current solution:**
* The most loading time consuming tasks are the application of GameObject components and hierarchies from Pre-cached VOBs and the Prefabs. This is heavily used already and consumes most of the time as we apply it to 11k objects when loaded.
* Precaching VOBs is a huge complexity addition where we now have 3 required phases of Component additions to VOBs
* The more we put into the cache, the more likely we will miss to keep track of when a cache needs to be altered or not. (Cache version flag)
* Loading fifteen thousand objects when we will see only a handful of them during game sessions feels so wrong

## Next steps:
* The Branch behind this PR can be referenced for a couple of weeks if needed
* A possibility could be to only cache TextureArray information and Meshes. No VOB object structure, but instead applying the meshes at the time of world loading.
* Some smaller improvements (like disabling meshes during loading time) are currently adopted to main #154

PR #148, opened 2024-10-02
Feature: texture animation

JucanAndreiDaniel:
## Description
Adds texture animation for materials that support it, mostly water chunks

## TODO:
- [ ] find a better place for TextureAnimator.I.Begin() to be called (needs to be called after the texture cache does it's thing)

## Out of Scope
We're not handling any mesh/material not used for the texture cache


JucanAndreiDaniel:
Superseeded by #150

PR #154, opened 2024-10-04
Feature/75 async loading based on device refresh rate

### LINKED ISSUE #75: Optimize world loading times based on FPS of current device
We currently load world chunks, VOBs, and NPCs in a pre-defined chunk size per frame. (e.g. vobsPerFrame=10). We could also think of a mechanism to improve:  
_Load only as many elements to keep constant FPS size for the current VR Device - do not have too low performance and therefore remove nausea_

I thought about a fps calculator which is monitoring fps during loading time and reports it to the actual async-loader if they need to reduce/raise their amount of calculations per frame.

We need to think about two topics:
1. fps can be calculated every frame (but should be smoothened to get an average over time) - Time.deltaTime 
2. VR devices might report a different refresh rate if the performance is too low (e.,g. half rates) - TryGetDisplayRefreshRate()

Hints:
* Whoever tests it should do it with an MVR device (to have two tests, one with the Editor+PC, one with the actual mobile hardware)
* VR RefreshRate thread: https://forum.unity.com/threads/how-do-you-get-the-device-refresh-rate-in-openxr.1056899/
* working FPSCalculator script with deltaTime:
    <details><summary>Click to expand</summary>
    
    ```c#
    using System.Collections.Generic;
    using TMPro;
    using UnityEngine;
    
    namespace GUZ.Core.Debugging
    {
        public class FpsCounter : MonoBehaviour
        {
            [SerializeField] private TMP_Text _label;
    
    
            // The number of seconds over which to average the FPS
            private float _averagingPeriod = 2.0f;
    
            // Interval in seconds between average FPS calculations
            private float _updateInterval = .5f;
    
            // Queue to store frame times
            private Queue<float> _frameTimes = new();
    
            // The sum of all frame times in the queue
            private float _totalTime;
    
            // The average FPS
            private float _averageFPS;
    
            // Timer to control the update frequency
            private float _updateTimer;
    
            void Update()
            {
                // Get the time taken for the last frame
                var frameTime = Time.deltaTime;
    
                // Add the frame time to the queue
                _frameTimes.Enqueue(frameTime);
                _totalTime += frameTime;
    
                // Remove old frame times that exceed the averaging period
                while (_totalTime > _averagingPeriod)
                {
                    _totalTime -= _frameTimes.Dequeue();
                }
    
                // Update the timer
                _updateTimer += Time.deltaTime;
    
                // Calculate and display the average FPS at the specified interval
                if (_updateTimer >= _updateInterval)
                {
                    // Calculate the average FPS
                    if (_frameTimes.Count > 0)
                    {
                        _averageFPS = _frameTimes.Count / _totalTime;
                    }
                    else
                    {
                        _averageFPS = 0.0f;
                    }
    
                    // Display the average FPS (for debugging purposes)
                    _label.text = "Average FPS: " + _averageFPS;
                    Debug.Log("Average FPS: " + _averageFPS);
                    // Reset the update timer
                    _updateTimer = 0.0f;
                }
            }
        }
    }
    
    ```
    </details>

### LINKED ISSUE #131: Loading performance: Disable Mesh + VOBs during loading
## Description
When we currently load the game, all meshes of world and VOBs will be loaded fully and disabled once culling kicks in, which is _after_ everything is loaded. On VR devices loading starts to stutter more and more over time. Therefore let's disable all meshes/GOs once they're loaded immediately to save performance during loading.
Image

## Tasks
- [ ] World + VOBs - Once object is loaded, call .SetActive(false)
- [ ] Once everything is loaded and normal Culling kicks in (e.g. VobMeshCulling) we might need to do an initial .Query() to re-enable elements next to us. (Based on experience, culling won't trigger an initial "visible" Event)

JaXt0r:
## Description
Before this branch, we hard coded amount of world chunks + VOBs being created per frame. Now we dynamically decide based on device's fps settings when to skip to next frame. We also left all meshes visible during loading time which increased stress on rendering device. Meshes are all disabled now until game is fully loaded.

## To test
* Start Game on a MVR device (preferrably Quest 2)
* Check how fast the game loads the world
  * loading should be faster or at least comparable to loading from main branch
  * loading should have less hickups during loading

## My own test results:
* New logic
  * 72 fps on Quest 2 where we skip a frame after 50% threshold -> Game loaded in 14 sec
* Old logic:
  * Game loaded in 13 sec

Hint: Quest 2 is our lower end device. On a newer device, the game will load even faster as the loading chunks per frame are dynamically calculated


JucanAndreiDaniel (approved):
from stuttery mess to 2 small stutters and the big one when loading is finished (no npc included)

PR #155, opened 2024-10-05
Feature/95 daedalus active perceptions - Assesslayer + AssessTalk

### LINKED ISSUE #95: Daedalus Active Perceptions
## Background:
When we start the game, Diego normally starts talking to us. Right now we need to click him to start the dialog.

## TODOs:
- [x] Implement perception logic (Perceptions are checked regularly (e.g. every ZS_*_LOOP() execution or after some timeouts?) -> #155
- [ ] There are 5 active perceptions which we should implement together. As we have no NPC interactions so far, assessplayer + body + item will be the ones to test.
    - [x] PERC_ASSESSPLAYER -> #155
    - [ ] PERC_ASSESSBODY - #162
    - [ ] PERC_ASSESSITEM - #161
    - [ ] PERC_ASSESSENEMY - #160
    - [ ] PERC_ASSESSFIGHTER - #159

## Technical details:
**From OpenGothic:**  
* Function where the elements are called: [Npc::perceptionProcess](https://github.com/Try/OpenGothic/blob/master/game/world/objects/npc.cpp#L3792)
* So when there's a .tick() call in OG, it checks for perception functions and executes them. I don't know when exactly, but I think when a ZS_LOOP is done. Try has a nextPerceptionTime timer, but I need to understand it better.
* Because Diego talks to us "after" he rotated his head towards us. Therefore I think ZS will be finished, then all perceptions executed and then ZS_* re-executed if no important dialogs are needed.

**From Daedalus:**  
* The functions to call on Daedalus are defined here: `Npc_PercEnable(self, PERC_ASSESSPLAYER, B_AssessSC);`
* So the NPC is basically just checking for the .important elements and then calls Ai_StartState("ZS_TALK"). So the NPC is executing the same function which is also used when an NPC talks to another NPC. 😄

image
image

### LINKED ISSUE #157: We need to call passive Perception Perc_ASSESSTALK rather than starting the dialog this way.
Where:
 VRNPC.cs

What:
                // FIXME - We need to call passive Perception Perc_ASSESSTALK rather than starting the dialog this way.
                DialogManager.StartDialog(gameObject, GetComponent<NpcProperties>(), true);

Why:
Daedalus will loop through all the nitty gritty details of an NPCs pre-behaviour before setting them into state ZS_TALK

JaXt0r:
## Done
* Perception logic is implemented inside AI behaviour
* Perc_Assessplayer is implemented

## To test
* GameConfiguration: SpawnNpcs=true
    * NpcIds
        * 1 - Diego
        * 201 - Jackal
        * 233 - Bloodwyn
        * 536 - Kyle
* Start game
* (new) Diego will start talking to us on his own
* When clicking "END" is Chapter 2 starting?

## Out-of-scope
* When NPCs start talking to us, we aren't locked in place as of now
* Other active perceptions aren't implemented so far and should be added once a game scenario is created where they apply.
* Followed by: #157

### DIFF
--- a/Assets/UnZENity-Core/Resources/Prefabs/Vobs/oCNpc.prefab
+++ b/Assets/UnZENity-Core/Resources/Prefabs/Vobs/oCNpc.prefab
@@ -28,14 +28,14 @@ Transform:
   m_PrefabAsset: {fileID: 0}
   m_GameObject: {fileID: 2202926895728732233}
   serializedVersion: 2
-  m_LocalRotation: {x: 0.0044286796, y: -0.46735078, z: 0.0030868165, w: 0.8840555}
-  m_LocalPosition: {x: -65.10339, y: 8.841797, z: 35.572765}
+  m_LocalRotation: {x: 0, y: 0, z: 0, w: 1}
+  m_LocalPosition: {x: 0, y: 0, z: 0}
   m_LocalScale: {x: 1, y: 1, z: 1}
   m_ConstrainProportionsScale: 0
   m_Children:
   - {fileID: 8679908380237941234}
   m_Father: {fileID: 0}
-  m_LocalEulerAnglesHint: {x: 0.614, y: -55.725, z: 0.076}
+  m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
 --- !u!114 &4778879049934742964
 MonoBehaviour:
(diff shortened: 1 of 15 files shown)


BoroBongo (approved):
Let's get it! Awesome stuff.

PR #156, opened 2024-10-05
fix: Npc armor adjustments for SoftSkinMeshes re-enabled.

JaXt0r:
## Description
1. NPC heads are back onto their shoulders when wering an armor.
2. Some NPC armors were broken. This was due to TextureArray creation removed cached Textures from Armors (e.g. Orry's armr texture also existed as VOB hanging around in world. Unity's TextureArray data copy removed it from memory and we got a broken Orry)

## To test
* Simply load world with all NPCs active and check a few if
   * Heads and
   * Armor are correct


BoroBongo (approved):
I love it! Awesome work!

PR #158, opened 2024-10-09
ugly hotfix Continuous perceptions (even after Diego dialog)

BoroBongo:
Without these changes (mostly AIVAR sets and call for ReEnableNpcs) After dialog with Diego other NPCs don't initiate the conversation if it's marked important (i.e. Bloodwyn).

To test:
- Spawn Diego (1), Skip (211), Thorus (200) and Bloodwyn (233)
- Spawn at GameStart
- Diego should start the conversation and proceed as normal in the game.
- Chapter 1 has started
- Walk/Teleport to OC
- Approach Bloodwyn and he should start the conversation with you about his pocket money :D

IMPORTANT: Don't talk to Diego while he's busy with walking back to the Camp!!!! There's something wrong happening with the routine/animation queuing and it will basically softlock any dialogs for the session.

Also not sure if related to this change but:
If we're in a dialog window with an NPC, let's say Thorus, and we pick one of the dialog options.. we need to wait until Thorus finishes his current animation (Wait for him to finish looking around) and only then the hero actually starts the dialog.

### DIFF
--- a/Assets/UnZENity-Core/Scripts/Globals/Constants.cs
+++ b/Assets/UnZENity-Core/Scripts/Globals/Constants.cs
@@ -129,6 +129,7 @@ namespace GUZ.Core.Globals
 
         public static class DaedalusConst
         {
+            public static int AIVInvincibleKey => GameData.GothicVm.GetSymbolByName("AIV_INVINCIBLE").GetInt(0);
             public static int AIVItemStatusKey => GameData.GothicVm.GetSymbolByName("AIV_ITEMSTATUS").GetInt(0);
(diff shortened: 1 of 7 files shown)


BoroBongo:
I used [NpcMeshCullingManager.cs](https://github.com/Gothic-UnZENity-Project/Gothic-UnZENity/pull/158/files#diff-91fd875d75a64a06d3344f689094e7caf6ef91637d27d70860b91bc13601a997) as the means of calling ReEnableNPCs as that was the only example that uses ReEnableNPCs - it was just convenient to set it statically there for testing. Please suggest a better place for it to live (assuming that the ReEnableNPC approach is correct)

JaXt0r (approved):
Merging into other branch to work on.

PR #165, opened 2024-10-12
Feature/133 load monsters

### LINKED ISSUE #133: Load monsters
> Forget about the description below. We now know, that we need to Alloc+Init every NpcInstance every time we call Wld_InsertNpc(). Please check #165 for details how we implemented it.

-----

## Description
We currently load NPCs only. To spawn them there are two Daedalus VM calls involved:
1. AllocateInstance() - Allocate memory and fetch empty instance with Index (needed for caching on Unity lookup)
2. and InitInstance() - Call member functions of instance (e.g. Mdl_SetVisual()) and fill up the object

Known issues:
1. The current cache (NpcCache) is optimized for single instance NPCs only. But monster instances spawn multiple times.
2. We only need to Alloc+Init an NPC/Monster instance once! Once done we need to copy over the data to a Vob for NPCs and work on.

## Tasks
**Loading/Spawning**
- [ ] We Alloc+Init a NpcInstance (Daedalus: INSTANCE *) only once
- [ ] When we create Unity objects based on this data, we need to copy NpcInstance data into Npc extends VirtualObject classes
  * Hint: These Npc objects are also the objects we use when we save the game)
  * So basically: Alloc+Init is only once per INSTANCE * to fetch basic data. All changes are handled within Npc VOB objects

**Further handling**
- [ ] Check if their idle animations are called automatically (perception and walking around)
- [ ] Ensure they're culled in and out like normal NPCs

JaXt0r:


JaXt0r:
Checked with @JucanAndreiDaniel together on this. Fine to be merged now.

PR #166, opened 2024-10-12
Feature 96/dialog subtitles

BoroBongo:
## TODOs
- [x] Subtitles for NPCs should be visible above their head and are rotated towards us (As we can't move in the future it will be a fixed value when dialog starts)
- [ ] Subtitles for our hero
    - [ ] should be slightly below our view (golden cut or so)
    - [ ] They should have a delay when moving with us for convenience (you can reuse the Chapter switch screen logic - there's a movement delay implemented already) - [HVRIntroduceChapter.cs](https://github.com/Gothic-UnZENity-Project/Gothic-UnZENity/blob/main/Assets/Gothic-UnZENity-HVR/Scripts/Components/HVRIntroduceChapter.cs)
- [ ] Dialog settings can be altered on our MainMenu
- [ ] (Bonus) We can also read and write the subtitle setting from gothic.ini. The ini file is already read in our game. We just need to implement the save as well. - Please read+save the settings inside PlayerPrefsManager then

### DIFF
--- a/Assets/UnZENity-Core/Scripts/Adapter/IInteractionAdapter.cs
+++ b/Assets/UnZENity-Core/Scripts/Adapter/IInteractionAdapter.cs
@@ -9,6 +9,7 @@ namespace GUZ.Core.Adapter
         float GetFrameRate();
         GameObject CreatePlayerController(Scene scene, Vector3 position = default, Quaternion rotation = default);
         void CreateVRDeviceSimulator();
+        void LockPlayerInPlaceCutscene();
         void LockPlayerInPlace();
         void UnlockPlayer();
         void TeleportPlayerTo(Vector3 position, Quaternion rotation = default);
--- a/Assets/UnZENity-Core/Scripts/Adapter/IInteractionAdapter.cs
+++ b/Assets/UnZENity-Core/Scripts/Adapter/IInteractionAdapter.cs
@@ -9,7 +9,6 @@ namespace GUZ.Core.Adapter
         float GetFrameRate();
         GameObject CreatePlayerController(Scene scene, Vector3 position = default, Quaternion rotation = default);
         void CreateVRDeviceSimulator();
-        void LockPlayerInPlaceCutscene();
         void LockPlayerInPlace();
         void UnlockPlayer();
         void TeleportPlayerTo(Vector3 position, Quaternion rotation = default);
--- a/Assets/UnZENity-Core/Scripts/Globals/GameData.cs
+++ b/Assets/UnZENity-Core/Scripts/Globals/GameData.cs
@@ -39,6 +39,9 @@ namespace GUZ.Core.Globals
             public static List<InfoInstance> Instances = new();
             public static bool IsInDialog;
 
+            // I'm too stupid to make it work through vfs need someone to help here
+            public static CutsceneLibrary CutsceneLibrary = new($"{GameGlobals.Settings.Gothic1Path}\\_work\\DATA\\scripts\\Content\\CUTSCENE\\Ou.csl");
+
             public static int GestureCount;
 
             public static class CurrentDialog
--- a/Assets/UnZENity-Core/Scripts/Globals/GameData.cs
+++ b/Assets/UnZENity-Core/Scripts/Globals/GameData.cs
@@ -39,8 +39,7 @@ namespace GUZ.Core.Globals
             public static List<InfoInstance> Instances = new();
             public static bool IsInDialog;
 
-            // I'm too stupid to make it work through vfs need someone to help here
-            public static CutsceneLibrary CutsceneLibrary = new($"{GameGlobals.Settings.Gothic1Path}\\_work\\DATA\\scripts\\Content\\CUTSCENE\\Ou.csl");
+            public static CutsceneLibrary CutsceneLibrary;
 
             public static int GestureCount;
 
--- a/Assets/UnZENity-Core/Scripts/Manager/GUZBootstrapper.cs
+++ b/Assets/UnZENity-Core/Scripts/Manager/GUZBootstrapper.cs
@@ -119,7 +119,7 @@ namespace GUZ.Core.Manager
         
         private static void LoadSubtitles()
         {
-            GameData.Dialogs.CutsceneLibrary = new($"{GameGlobals.Settings.Gothic1Path}\\_work\\DATA\\scripts\\Content\\CUTSCENE\\Ou.csl");
+            GameData.Dialogs.CutsceneLibrary = new($"{GameContext.GameVersionAdapter.RootPath}\\_work\\DATA\\scripts\\Content\\CUTSCENE\\Ou.csl");
         }
 
         private static void LoadVideos()
--- a/Assets/UnZENity-Core/Scripts/Manager/GUZBootstrapper.cs
+++ b/Assets/UnZENity-Core/Scripts/Manager/GUZBootstrapper.cs
@@ -31,6 +31,7 @@ namespace GUZ.Core.Manager
             LoadGothicVm(gothicDir);
             SetLanguage();
             LoadDialogs();
+            LoadSubtitles();
             LoadVideos();
             LoadSfxVm(gothicDir);
             LoadPfxVm(gothicDir);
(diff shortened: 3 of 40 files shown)


Assets/UnZENity-Core/Scripts/Adapter/IInteractionAdapter.cs
```diff
@@ -9,6 +9,7 @@ public interface IInteractionAdapter
         float GetFrameRate();
         GameObject CreatePlayerController(Scene scene, Vector3 position = default, Quaternion rotation = default);
         void CreateVRDeviceSimulator();
+        void LockPlayerInPlaceCutscene();
```
JaXt0r: Why not reusing function below? In the end it seems like we call the same VrController.Lock() function. Feel free to reuse until you need special treatment.

Assets/UnZENity-Core/Scripts/Npc/Actions/AnimationActions/Output.cs
```diff
@@ -46,7 +51,24 @@ public override void Start()
...
                 // FIXME - Show subtitles above NPC
+                PrintDialog();
+            }
+        }
+
+        private void PrintDialog()
+        {
+            // FIXME - Show subtitles somewhere next to Hero (== ourself/main camera)
+            var currentMessage = GameData.Dialogs.CutsceneLibrary.Blocks.Find(x => x.Name == OutputName).Message;
+            NpcInstance foo = (NpcInstance)GameData.GothicVm.GlobalHero;
```
JaXt0r: foo? 😜

Assets/UnZENity-Core/Scripts/Npc/Actions/AnimationActions/Output.cs
```diff
@@ -46,7 +51,24 @@ public override void Start()
...
+            if(_isHeroSpeaking){
+                Debug.Log($"PrintDialog: {{ Hero: {currentMessage.Text}}}");
```
JaXt0r: I try to reduce log statements to a minimum in normal game play. My normal mode is:
1. During develepment I add Debug.Log() statements
2. Once I create the PR, I remove them again
3. If there are bugs in the future, I check via Debugger (if any)
I think you can safely do this as well. (Especially as the actual text is printed inside the Adapter the next call)

Assets/UnZENity-VR/Scripts/Adapter/VRSubtitlesAdapter.cs
JaXt0r, line 1: Thanks for putting the logic into the Adapters already.
As a recommendation, I'd say let's put this Adapter logic into the DialogAdapter as Subtitles are part of the Dialog.

Assets/UnZENity-VR/Scripts/Components/UI/VRDialog.cs
```diff
@@ -49,6 +49,12 @@ public void ShowDialog(GameObject npcGo)
             hoverElements.AddRange(_dialogItems.Select(i => i.GetComponentInChildren<TMP_Text>().gameObject).ToList());
             _uiEventsHandler.SetElementsToHover(hoverElements, true);
             
+            StartCoroutine(ShowDialogWithDelay());
+        }
+
+        private System.Collections.IEnumerator ShowDialogWithDelay()
+        {
+            yield return new WaitForEndOfFrame();
```
JaXt0r, line 58: Smart. This is the one, that the dialog rotation isn't shown in game?
Please add a comment why this call is done.

Assets/UnZENity-Core/Scripts/Globals/GameData.cs
```diff
@@ -39,6 +39,9 @@ public static class Dialogs
             public static List<InfoInstance> Instances = new();
             public static bool IsInDialog;
 
+            // I'm too stupid to make it work through vfs need someone to help here
+            public static CutsceneLibrary CutsceneLibrary = new($"{GameGlobals.Settings.Gothic1Path}\\_work\\DATA\\scripts\\Content\\CUTSCENE\\Ou.csl");
```
JaXt0r: The lookup is fine this way. Just put it inside GuzBootstrapper please. We already have a function where we load all VFS and (e.g.) the GothicVM. Just add another method like LoadCutscenes (Or better LoadSubtitles? ;-) ).

JaXt0r (changes requested):
Nice feature. I really like you understood all the magic with Adapter + GameData + ...
Please check my code improvement comments.

JaXt0r (commented):
I think you added a second Audio Listener to the additional camera. Please check if you can remove it as it clutters the log.

image

JaXt0r (commented):
image

It seems like all the CutsceneLibrary entries can't be parsed fully. Can you briefly check in with Luis if this is ok? If yes, we just do these warnings to > /dev/null, of not, maybe he can provide some help.

JaXt0r (commented):
image

Something is wrong when spawning Diego only.

This line of code crashes:
```c#
                // We need to look through all RootGOs and fetch the first matching HVRDialog Component.
                _subtitlesComponent = scene.GetRootGameObjects()
                    .Select(i => i.GetComponentInChildren<VRSubtitles>())
                    .First(i => i != null);
```

Assets/UnZENity-Core/Scripts/Npc/Actions/AnimationActions/Output.cs
BoroBongo: Ahh forgot to change that haha

Assets/UnZENity-Core/Scripts/Adapter/IInteractionAdapter.cs
BoroBongo: This is different treatment as in the LockPlayerInPlace method it locks both transform and rotation. We want to rotation to stay so player can look around freely

JaXt0r (commented):
image

Some .meta files are missing. We need to commit them. Otherwise Scenes/Prefabs can't correlate .cs Components with their appropriate .cs files.

Assets/UnZENity-Core/Scripts/Globals/GameData.cs
BoroBongo: ZenKit goes with calling it Cutscenes but I agree that we can call it LoadSubtitles here but keep the current naming in deeper levels of scope. 🫡

Assets/UnZENity-Core/Scripts/Adapter/IInteractionAdapter.cs
JaXt0r: But in VR, I can move my head around normally. Isn't it the same as if I could rotate via Thumbstick? This will affect rotation as well, correct?

BoroBongo: Ah! You're correct - this is for the Player rotation and not camera.

Assets/UnZENity-VR/Scripts/Adapter/VRSubtitlesAdapter.cs
BoroBongo: I think keeping them separate for now will be a better choice. They point towards and manipulate two different prefabs. I also like the distinction between them - we contain all 'game' logic (like choosing the dialog option) separate from the basic UI element which the subtitles are.

JaXt0r (approved):
Nicely done. Go on in merging it.

Assets/UnZENity-Core/Scripts/Manager/GUZBootstrapper.cs
```diff
@@ -116,6 +117,11 @@ private static void LoadDialogs()
             infoInstances.ForEach(i => GameData.Dialogs.Instances.Add(i));
         }
         
+        private static void LoadSubtitles()
+        {
+            GameData.Dialogs.CutsceneLibrary = new($"{GameGlobals.Settings.Gothic1Path}\\_work\\DATA\\scripts\\Content\\CUTSCENE\\Ou.csl");
```
JucanAndreiDaniel: Can we change this into something like `GameGlobals.Settings.GothicPath`? As to not have g1 hardcoded

JucanAndreiDaniel: Forgot about the adapter path

```suggestion
            GameData.Dialogs.CutsceneLibrary = new($"{GameContext.GameVersionAdapter.RootPath}\\_work\\DATA\\scripts\\Content\\CUTSCENE\\Ou.csl");
```

JucanAndreiDaniel (changes requested):

Assets/UnZENity-Core/Scripts/Manager/GUZBootstrapper.cs
BoroBongo: Great catch! Thanks

JucanAndreiDaniel (approved):

BoroBongo:
This is the first V1 of Subtitles change, not resolving the issue yet as there are still some things left to do.

PR #168, opened 2024-10-15
attempt at a fix - initialising lights before worldCreation

BoroBongo:
To test:
Fly around in the editor - loaded world meshes shouldn't be missing (no holes)

### DIFF
--- a/Assets/UnZENity-Core/Scripts/Vob/StationaryLight.cs
+++ b/Assets/UnZENity-Core/Scripts/Vob/StationaryLight.cs
@@ -1,6 +1,9 @@
 using System;
 using System.Collections.Generic;
+using System.Threading.Tasks;
 using GUZ.Core.Extensions;
+using GUZ.Core.Manager;
+using GUZ.Core.Util;
 using UnityEngine;
 using UnityEngine.Profiling;
 
@@ -146,11 +149,8 @@ namespace GUZ.Core
 
         private void Awake()
         {
-            if (!Lights.Contains(this))
-            {
-                Lights.Add(this);
-                _threadSafeLightData.Add((transform.position, Range));
-            }
+            Lights.Add(this);
+            _threadSafeLightData.Add((transform.position, Range));
         }
 
         private void OnDestroy()
@@ -215,7 +215,11 @@ namespace GUZ.Core
             Shader.SetGlobalVectorArray(GlobalStationaryLightPositionsAndAttenuationShaderId,
                 lightPositionsAndAttenuation);
             Shader.SetGlobalVectorArray(GlobalStationaryLightColorsShaderId, lightColors);
-            _threadSafeLightData.Clear(); // Clear the thread safe data as it is no longer needed.
+        }
+
+        public static void ClearThreadSafeData()
+        {
+            _threadSafeLightData.Clear();
         }
 
         public static int CountLightsInBounds(Bounds bounds)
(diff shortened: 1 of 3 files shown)


Assets/UnZENity-Core/Scripts/Vob/StationaryLight.cs
```diff
@@ -243,5 +250,34 @@ private void GatherRenderers()
...
+
+        public static async Task InitializeThreadSafeLightData()
+        {
+            Lights.Clear();
+            _threadSafeLightData.Clear();
+
+            // Find all StationaryLight components, including those on inactive GameObjects
+            var allLights = Resources.FindObjectsOfTypeAll<StationaryLight>(); // ~350 for G1
+
+            var lightsPerFrame = 30;
```
JaXt0r: Why introducing this property? The FrameSkipper is taking care of checking when enough processing time is passed automatically. Or did you see some impact on it when usin the FrameSkipper after every call?

If you unsure how many time the FrameSkipper will _not_ skip during a frame, you can easily re-enable the debug lines (__debugSkippedCount_) and check the logs with and without your 30x pre-condition.

Assets/UnZENity-Core/Scripts/Vob/StationaryLight.cs
```diff
@@ -212,7 +215,11 @@ public static void InitStationaryLights()
             Shader.SetGlobalVectorArray(GlobalStationaryLightPositionsAndAttenuationShaderId,
                 lightPositionsAndAttenuation);
             Shader.SetGlobalVectorArray(GlobalStationaryLightColorsShaderId, lightColors);
-            _threadSafeLightData.Clear(); // Clear the thread safe data as it is no longer needed.
+        }
+
+        public static void ClearThreadSafeData()
```
JaXt0r: This function never gets called. Can you (e.g.) add it to InitStationaryLights? (Or simply clear the _threadSafeLightData there.)

Assets/UnZENity-Core/Scripts/Vob/StationaryLight.cs
```diff
@@ -243,5 +250,34 @@ private void GatherRenderers()
                 }
             }
         }
+
+        public static async Task InitializeThreadSafeLightData()
```
JaXt0r: This static function currently resides in a MonoBehaviour. Please consider moving it to the StationaryLightsManager as it's a static singleton function which is more aligned to our coding behaviour on a Manager, rather than a Component.

JaXt0r (changes requested):
It works. Nicely done. And it's super fast (even on my machine in Editor Mode it won't take longer than a 2-4 seconds). Please check a few comments below.

Assets/UnZENity-Core/Scripts/Vob/StationaryLight.cs
BoroBongo: We can't clear it until World gets created - otherwise _StationaryLightCount inside Lit-World shader (which world chunks are using) won't be properly initialised. I'm calling this right after `WorldCreator.CreateAsync()`
```csharp
                // 4.
                if (config.EnableWorldMesh)
                {
                    // initialize Lights before world creation
                    vobRoot.SetActive(true); // temporary enable vobRoot
                    await StationaryLight.InitializeThreadSafeLightData();
                    vobRoot.SetActive(false); // disable to save some seconds in loading time ;p

                    await WorldCreator.CreateAsync(config, GameGlobals.Loading, worldRoot);
                    StationaryLight.ClearThreadSafeData(); // <-- here
                }
```

Assets/UnZENity-Core/Scripts/Vob/StationaryLight.cs
BoroBongo: Good shout! Requested change applied 🤠

Assets/UnZENity-Core/Scripts/Vob/StationaryLight.cs
BoroBongo: Yup these checks aren't needed. I initially misunderstood the concept of target frame rate. Requested changes applied

JaXt0r (approved):
👍

PR #169, opened 2024-10-18
Feature/163 savegame loading v2

JaXt0r:
## Background
NPCs are now spawned completely from a save game.
This includes:
1. NPCs which were visible at the time of saving the game
2. Not visibile ones (Both are handled separately inside Gothic SaveGame files)

## To test
- [ ] Create a few save games with some NPCs visible and some not
- [ ] When game is loaded, check if you can see them at the correct spot and with correct routine
- [ ] Example: Diego - he changes routine (look) after you talked to him (walk) and once invisible he is re-spawned from SaveGame at Old Camp
- [ ] But please test with some other NPCs as well

PR #170, opened 2024-10-19
Feature/30 savegame saving v1

### LINKED ISSUE #30: Savegame saving - v1
> I recommend picking this up after: #28 

Our game only supports _New Game_ at the moment. We want to support official G1 savegames. As ZenKit supports save&reload, let's leverage it to have G1 compatible save games for Gothic-UnZENity.

Links:
* [SaveGame.cs](https://github.com/GothicKit/ZenKitCS/blob/main/ZenKit/SaveGame.cs) at ZenKit

## TODOs

- [x] Create a Gothic save slot where you changed some stuff. (e.g. moved item, killed NPC)
- [x] Implement test to save a current state during gameplay into a save game
- [x] Check If the saved assets match expectations when loaded
    - [x] into G1 classic
    - [x] into Gothic-UnZENity
- [x] Implement Save button inside game menu

## Test tasks
- [x] Create a Component to compare two save games. It should loop through the VOBs (etc.) and compare order of elements and also value of Vob.Properties.
- [x] Test it via: Create an original G1 Save game, load it into UnZENity, save it to another SaveGame slot, and compare values.

## Hints:
* I figured out, that G1 save games store NPCs inside Vobs[].Type=oCNPC.
* It also stores NPCs/Monsters only! if they are nearby hero during save time.
* Also if we load a G1 save game, walk away from an NPC and save again, the updated save game won't include the far away NPC

JaXt0r:
> ❗ HINT: Can be reviewed and merged after #169

## To test
* Save and load the game in multiple ways during a session
  * Is hero spawning correct?
  * NPCs doing their job?
  * Are Items where we left them before saving+loading?

## Out-of-Scope/Known issues
* SaveGame menu is always active. So you could potentially hit _Save_ before entering a game and crash it. ;-)
* Thumbnail of an UnZENity save contains the MainMenu. Something to ignore when creating a screenshot later.


JaXt0r:
I need this branch for future implementations. I tested it a few times during development. As it's an _add -on_ feature, it's safe to accept.

PR #206, opened 2025-02-10
Feature/143 eat items

JaXt0r:
You can now grab food/potions and put it to your mouth. Then the drink/eat sound is played and the object gets destroyed once done.

## To test
* Go into Gomez' room
* Eat all that stuff from his table

## Boundary
* It's currently, that the mesh gets destroyed only. No inventory handling or removing the object from a save game so far


JaXt0r:
Tested and works. We can't rotate items in our hand these days. But it's unrelated to the actual eating feature.

PR #213, opened 2025-02-16
Add case insensitive file search

JucanAndreiDaniel:
Fixes ini and cache files loading on case sensitive filesystems


JaXt0r (approved):
Tested on Windows: Works

PR #217, opened 2025-02-24
Fix Level change triggers

JucanAndreiDaniel:
Reintroduce working level change triggers, they were broken since a refactor of the game manager.


JaXt0r (approved):
Works. Small update: Good that you mentioned it. TextureArray doesn't need to be recreated every time.

PR #223, opened 2025-03-28
VR Seat working again

JucanAndreiDaniel:
Spawnpoint with a seat right near it: OCR_OUTSIDE_HUT_2

please make sure the fade also works properly


JaXt0r (approved):
Code looks good. Didn't test so far, but will do later. Feel free to merge already.

PS: If you do a short recording and post it in our channel, this would be great to see for our gamers.

Sitting next to Mud and chilling like a boss. 😆

JaXt0r:
Tested now.
Findings:
1. I need to tap the object before the indicator for "seat now" appears. Can we have a ray caster from a distance to sit already? Like when we start an NPC dialog with ray caster.
2. Fading when moving to the spot is white. Can we have it in black?
3. Teleport is still active when we aren't allowed to walk. (Could be fixed globally with "DisableWalking" and disable the teleport feature there too.)

PR #228, opened 2025-04-16
Random Fixes

JucanAndreiDaniel:
Random fixes accumulated from different branches and testing


JaXt0r (approved):
Thanks for these G2 fixes.

PR #229, opened 2025-04-16
Perceptions and attitudes

JucanAndreiDaniel:
Implement some more Npc needed stuff to make them seem more alive/interactive/aware

### DIFF
--- a/Assets/UnZENity-Core/Scripts/Caches/StaticCache/StationaryLightCacheCreator.cs
+++ b/Assets/UnZENity-Core/Scripts/Caches/StaticCache/StationaryLightCacheCreator.cs
@@ -62,8 +62,15 @@ namespace GUZ.Core.Caches.StaticCache
                 {
                     var fire = (Fire)vob;
 
-                    if (fire.VobTree.IsNullOrEmpty() || (GameContext.GameVersionAdapter.Version == GameVersion.Gothic2 && fire.VobTree == "FIRETREE_LARGE.ZEN"))
+                    if (fire.VobTree.IsNullOrEmpty())
                     {
+                        continue;
+                    }
+
+                    // FIXME - For some reason, FIRETREE_LARGE.ZEN is broken in G2. Let's fix it properly later.)
+                    if (GameContext.GameVersionAdapter.Version == GameVersion.Gothic2 && fire.VobTree == "FIRETREE_LARGE.ZEN")
+                    {
+                        Debug.LogError("For some reason, FIRETREE_LARGE.ZEN is broken in G2. Let's fix it properly for caching.");
                         continue;
                     }
 
--- a/Assets/UnZENity-Core/Scripts/Caches/StaticCache/StationaryLightCacheCreator.cs
+++ b/Assets/UnZENity-Core/Scripts/Caches/StaticCache/StationaryLightCacheCreator.cs
@@ -5,6 +5,7 @@ using GUZ.Core.Manager;
 using GUZ.Core.Util;
 using MyBox;
 using UnityEngine;
+using ZenKit;
 using ZenKit.Vobs;
 using Light = ZenKit.Vobs.Light;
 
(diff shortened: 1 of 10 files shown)


Assets/UnZENity-Core/Scripts/Caches/StaticCache/StationaryLightCacheCreator.cs
```diff
@@ -61,7 +62,7 @@ public async Task CalculateStationaryLights(List<IVirtualObject> vobs, Vector3 p
                 {
                     var fire = (Fire)vob;
 
-                    if (fire.VobTree.IsNullOrEmpty())
+                    if (fire.VobTree.IsNullOrEmpty() || (GameContext.GameVersionAdapter.Version == GameVersion.Gothic2 && fire.VobTree == "FIRETREE_LARGE.ZEN"))
```
JaXt0r: Please add a comment why this workaround is needed for G2.

Assets/UnZENity-Core/Scripts/Manager/NpcHelper.cs
```diff
@@ -227,5 +250,14 @@ private static NpcProperties2 GetProperties([CanBeNull] NpcInstance npc)
         {
             return npc?.GetUserData2().Props;
         }
+
+        public static bool CanSenseNpc(NpcInstance self, NpcInstance other, bool freeLOS)
```
JaXt0r, line 250: Shouldn't we add a Parameter to sway "which sense is checked? As senses are hear, see, smell and have different calculations?

Assets/UnZENity-Core/Scripts/_Npc2/NpcAiManager2.cs
```diff
@@ -341,8 +360,63 @@ public bool ExtNpcOwnedByNpc(ItemInstance item, NpcInstance npc)
...
+        public void UpdateEnemyNpc(NpcInstance self)
+        {
+            var selfNpc = self.GetUserData2();
+            var selfPosition = selfNpc.Go.transform.position;
+            var enemyNpc = selfNpc.Props.EnemyNpc;
+
+            NpcContainer2 foundNpc = null;
+            if (enemyNpc != null && NpcHelper.CanSenseNpc(self, enemyNpc, true))
+            {
+                foundNpc = enemyNpc.GetUserData2();
```
JaXt0r, line 408: You are reassigning this value via LINQ later again. Why not returning the value here?

Assets/UnZENity-Core/Scripts/_Npc2/NpcManager2.cs
```diff
@@ -121,7 +121,7 @@ public void ExtCreateInvItems(NpcInstance npc, uint itemId, int amount)
         {
             // We also initialize NPCs inside Daedalus when we load a save game. It's needed as some data isn't stored on save games.
             // But e.g. inventory items will be skipped as they are stored inside save game VOBs.
-            if (!GameGlobals.SaveGame.IsWorldLoadedForTheFirstTime)
+            if (!GameGlobals.SaveGame.IsWorldLoadedForTheFirstTime || npc.GetUserData2() == null)
```
JaXt0r: Please add a comment when this NPC==null happens.

Assets/UnZENity-Core/Scripts/_Npc2/NpcProperties2.cs
```diff
@@ -83,5 +86,11 @@ public enum LoopState
         public float StateTime;
         public LoopState CurrentLoopState = LoopState.None;
         public AbstractAnimationAction CurrentAction;
+        
+        // Attitudes
+        // HINT: These values are only used when checking the attitude towards the player
+        // HINT: for attitudes between NPC we directly use the guild attitude
```
JaXt0r, line 92: Thanks for the comment. 👌

JaXt0r (changes requested):
A few comments requested only.

Assets/UnZENity-Core/Scripts/Manager/NpcHelper.cs
JucanAndreiDaniel: Yes indeed, we should have an object to say by which sense npc was detected, but as a prototype function to make sure the assessenemy perc works it's fine

Assets/UnZENity-Core/Scripts/_Npc2/NpcAiManager2.cs
JucanAndreiDaniel: After looking at the profiler, the LINQ was using too much cpu time and i've decided to just do manual checks

JaXt0r (approved):
I added some comments. Approved and Merged.

PR #237, opened 2025-04-28
feat: Refactor PlayTransitionAnimation

JucanAndreiDaniel:
Refactor PlayTransitionAnimation from UseItemToState to fix infinite recursion when orcs are eating food

# To test:
1. go to orc city to check that the game is no longer freezing from infinite recurssion
2. check if babe brush still works as intended


JaXt0r (approved):
Oh my. I wasn't aware, that Orcs are also eating in the game. :-D

image
image

Oh my. I tested also with main. Thanks for fixing this annoying infinite loop.

PR #257, opened 2025-05-31
Upgrade unity 6 - LTS

JaXt0r:
## To test
* Build Quest/Pico version
* Check if game boots and you can play the first minutes in game

Already tested:
* PCVR build + Play of .exe


JaXt0r:
Tested a few times by me and @BoroBongo

PR #265, opened 2025-06-08
Save&Load fixes - 2025-06

### LINKED ISSUE #254: NPC next to Orry sometimes trying to use the wheel...
... but in original G1, he's just standing there:

Image

JaXt0r:
## Fixes
- [x] AlignToWp is now calculated based on current NPC position, not just active WP. (Catches exceptions after save game loading for e.g. Guard (240))
- [x] Guard next to Orry won't try to open the gate when starting the game (technically speaking: oCMobInter are now always open, when the game starts)
  - [x] Buggy again. We need to find a smarter way to "ignore" oCMobInter/....State when G1 also ignores it. But not for all vob types.
- [x] When clicking on "New Game" while being in world
  - [x] No exception will be thrown.
  - [x] Diego will restart dialog after loading (NpcKnowsInfo is now fetched from SaveGame instead of separate List).
- [x] When loading a Save Game
  - [x] e.g. if Diego walked away while saving in G1, he will walk on in UnZENity (aka last known routine+state is going on)
- [ ] Create Save Game (SG) from New Game
  - [ ] Compare it with a normally created save game from G1
  - [ ] Take care, that Diego's new Routine is stored on SG after dialog
- [ ] Create Save Game from G1 and overwrite it in UnZENity
  - [ ] Everything should be the same, except: Position of NPCs nearby
- [x] Check if SoundDayTime and Sound still working (Moved init logic from Initializer to Start() when object should be enabled.
  - [x] Didn't work since LazyLoading of VOBs. Need to check with VobContainer logic as sub-elements like fire->sound not working.
- [x] Are Music Colliders still working?
- [x] When we load a game, we currently reinitialize Routines and overwrite Diego's routine after intro (Rtn_Start_1) with the initial one (Rtn_PreStart_1)


JaXt0r:
Next version of SaveGame support. We have about 300 diffs and an overall matching of lines for a SaveGame save, when we do it after Diego's dialog at the start. Will take a break from it and work on later.

PR #286, opened 2025-07-21
Feature/270 swimming and diving

JaXt0r:
To test:
* Walk into water when spawning into the game
* First you will be slower (knee deep)
* Then you will start to swim with bouling in water
* When both Grabbing buttons are clicked, you will dive under water
* Under water and on top, you will move by pulling both Grab buttons and then move your hands


JaXt0r:
I need to merge it as I have fixes in here, I want to reuse for other PRs (Fight).

PR #287, opened 2025-07-23
Monster and Ai_Attack() implementation

JaXt0r:
## To test
1. Spawn yourself next to Molerat in starting area
2. Grab the sword
3. Swing it -> You should hear a sound and vibrating of controller
4. Check if Molerat sees you and is warning, then attacking you
5. Then kill it with one strong swing!

Hint: Some AI_Attack() logic is implemented, but not yet fully. So expect bugs apart from the elements mentioned above.


JaXt0r:
Tested in Marvin Mode branch.

PR #312, opened 2025-10-08
Feature/misc

### LINKED ISSUE #270: Implement swimming v2
## Todos

- [x] Check when player is under water (Ray cast + check if water is around us?)
- [x] Alter fall down gravity when in water (slowly going up)
- [x] implement arm swimming. References:
  - [x] Reference game: [Subside](https://youtu.be/ZviYkfnoTkw?si=0KwAAhdjkQRMM2KT)
  - [x] Reference implementation: https://youtu.be/ViQzKZvYdgE?si=QW4ehbjX4XPt5x3C

Bonus (split into separate issues if needed):
- [x] Alter sight when under water (dive feeling)
- [x] When on the surface, do not dive under automatically. Only with a button click or certain force
- [x] Disable backpack (etc.) when in water
- [x] Bubbles around player (Let's see... Billboards might look awkward)
- [x] Implement air bar below player
- [x] Rotate player collider when in water --> 1.8m height to rotating 90° flat (otherwise we can't swim out in areas like Xardas old tower inside.) -> implementation: keep CharacterController, but raise steep offset to "hop" on top of water
- [x] Handle swim and dive grips separately per hand
- [x] Only handle swim movement of hand, when underwater.+

Bugfixing
- [x] synthetic rotation via thumbstick breaks arm swim
- [x] Shader change of objects (world mesh done, objects to follow)
- [x] Try with 3D bubbles. 2d ones look awkward in VR..
- [x] ~~Experiment with alpha of drowning values. 0.8f is quite bright. Maybe go with 0.9f?~~ -> 0.8f is fine. Otherwise it's too pitch black.

JaXt0r:


JaXt0r:
@jetbrains-junie the PR is missing a feature: When swimming, there can be two speeds for Walking and Running. Currently Running is disabled when Swimming. But it should be re-enabled and have a new speed (not the speed like running on solid ground).

PR #333, opened 2025-11-16
Valve Index Knuckle - finger tracking

JaXt0r:
Added XRHands package from Unity and activated this feature via VR Subsystem in Project Settings.

Tested:
✅ PCVR Index Controller -> all fingers are tracked correctly
✅ PCVR Quest 2 -> Trigger is causing three last fingers to move (as before)


JaXt0r:
Tested with Knuckles and also normal Quest2 controllers. Works like a charm.

PR #350, opened 2025-12-26
Feature/zombie in xardas tower

JaXt0r:
## Changes
* Fix Guild check so that e.g., a Zombie is attacking us correctly
* Health bar is visible on player
* Monsters/NPCs have a fist/hand collider to track fist attacks - dynamically updating its space with each frame (e.g., troll opening his hands for a slap)


JaXt0r:
Temp push as I'm working on another branch these days. The named fixes work, but the feature is still WIP.

PR #353, opened 2026-01-02
Pipeline/custom keystore

JaXt0r:


JaXt0r:
Tested and it works now. Issue was, that the Keystore information need to be provided within the custom C# builder class and not only on YAML CI configuration.

PR #359, opened 2026-04-23
Update VRPlayerItemAdapter.cs

piotrb92:
only added a check if Rigidbody is not null before setting isKinematic to false 
This script assumed everything grabbable will have a rigidbody 
As a result, ladders were not working.
It was also conflicting with the climbing system that I was working on.


piotrb92:
@JaXt0r

JaXt0r (approved):

PR #368, opened 2026-06-13
Stack/01 fixes standalone

JucanAndreiDaniel:
## Summary
First in stack. Just some bugfixes for now.

## Changes
- **Linux world path separators** — `SaveGameService.ChangeWorld` normalises `\` to `/` before `Path.GetFileName`, which on Linux otherwise keeps the `AddonWorld\` prefix and fails to match the loaded world.
- **ChangeLevel trigger** — `oCTriggerChangeLevel` box collider now has `IsTrigger = 1` so level-change `OnTriggerEnter` actually fires.
- **Lockpick on unlocked containers** — `VRContainerDoorPickingInteraction` returns after deactivating itself, so it no longer starts the delayed coroutine on an already-unlocked container.
- **G2 cutscene suffix** — corrected from `CSL` to `LSC`. Gothic 2 english steam version has OU.LSC 

## How to test
- On Linux, change worlds (G2 portal to oldworld/addonworld) and confirm the target world tries to load (world traversal broken on current branch, should work on branch 04-culling).
- Walk into a ChangeLevel trigger and confirm the level transition starts.
- Interact with an already-unlocked container and confirm no lockpick minigame starts.


JaXt0r (approved):
Move on.

But there is no change with regards to lock picking as mentioned in the PR description:

```
Lockpick on unlocked containers — VRContainerDoorPickingInteraction returns after deactivating itself, so it no longer starts the delayed coroutine on an already-unlocked container.
```

JucanAndreiDaniel:
probably the change was already in main, and as you merged main into this branch, it disappeared
commit is still there - https://github.com/Gothic-Unity-Project/Gothic-Unity/pull/368/changes/76c6c02e7e72d881e05a830a8cc36d0504abb4b5

PR #369, opened 2026-06-13
Stack/02 vob loading

JucanAndreiDaniel:
## Summary
Reduces VOB lazy-init overhead and fixes bed VOBs spawning the wrong prefab.

## Changes
- **Queue dedup** — replaces the O(n) `Queue.Contains()` guard with a per-loader `IsQueued` flag; the init queue is batch-drained (10/frame), skips entries destroyed in the meantime, and is cleared on world change so stale entries from the previous world are dropped.
- **Bed prefab routing** — `oCMobDoor` VOBs whose visual name starts with `BED_` now resolve the `VobBed` prefab instead of `VobDoor`.

## How to test
- Load a populated world and confirm VOBs stream in without the previous Contains() hitch; Should load the vobs faster.
- go near a bed and check that you don't get spammed with errors regarding door locks

### DIFF
--- a/Assets/Gothic-Core/Scripts/Adapters/Scenes/WorldScene.cs
+++ b/Assets/Gothic-Core/Scripts/Adapters/Scenes/WorldScene.cs
@@ -165,6 +165,9 @@ namespace Gothic.Core.Adapters.Scenes
                     Logger.LogError(e.ToString(), LogCat.Loading);
                 }
 
+                // The culling sweep inside WorldSceneLoaded only _enqueued_ all VOBs around the spawn point for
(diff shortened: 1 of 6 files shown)


JaXt0r (approved):
private const int VobInitBatchSize = 10;
--> Does it mean we load max 10 vobs per frame?

JucanAndreiDaniel:
> ```
>     private const int VobInitBatchSize = 10;
> ```
> 
> --> Does it mean we load max 10 vobs per frame?

yes, but it's fast on quest as well
the reason for this thing is to load faster (before was sequential)

PR #370, opened 2026-06-13
Stack/03 animation playables

JucanAndreiDaniel:
## Summary
Rewrites the NPC animation system onto Unity's `AnimationScriptPlayable` with a Burst pose job sampling baked native pose buffers, replacing the old managed keyframe model. This is the foundation for the AI/combat work in later PRs.

## Changes
- **Model** — removes `AnimationKeyFrame`; adds `AnimationPoseJob` (Burst sample+blend over `NativeArray` pose data) and `AnimationSkeleton`.
- **Tracks** — `AnimationTrack` now owns shared native position/rotation buffers (with InvertYAxis` flag for sitting poses); `AnimationTrackInstance` is reduced to a playbackcursor.
- **Baking** — `AnimationService` bakes each `.man` once: pins root X/Z to zero, storeset, computes root-motion speed, and flags sitting poses.
- **Adapter** — `AnimationSystem` schedules the pose job through an `AnimationScriptPlayable` graph; root motion is applied via `RootCollisionHandler`.
- Editor `AnimationSystemWindowTool` updated for the new pipeline.

## How to test
- Spawn NPCs and confirm idle/walk/turn animations play and blend correctly.
- Verify sitting poses (bench/throne) are oriented correctly (Y inversion).
- Confirm root motion (walk translation) advances the NPC and no longer slides.

This is still not foolproof enough, there will still be issues, but the animation doesn't seem to be a bottleneck now on the quest


JaXt0r (approved):
Animations are working. Code looks good. I only got a few exceptions during play about Mobs and their usage. E.g., Diego is not sitting on his bench. But I don't expect it's related to the Animation rework.

PR #371, opened 2026-06-13
Stack/04 culling

JucanAndreiDaniel:
## Summary
Reworks distance culling to be band-based with hysteresis and lets NPC routines progress while meshes are unloaded.

## Changes
- **Shared base** — common bounding-sphere/bucket plumbing moved into `AbstractCullingDomain`/`AbstractCullingService`.
- **NPC culling** — distance bands with hysteresis (no thrashing at the boundary); visible NPCs tracked in a `HashSet`; routine waypoints advance even while the mesh is culled, so NPCs are in the right place when they re-appear.
- **VOB mesh culling** — three near-identical wrapper handlers collapsed into one per-bucket closure loop.
- **VOB sound culling** — `List` + `ToArray()` replaced by a reusable doubling array (no per-update allocation).

## How to test
- Walk toward/away from NPCs and confirm they load/unload at the band edges without flicker near the threshold.
- Leave an NPC's area, return after some time, and confirm it has progressed along its routine.
- Confirm VOB meshes and positional sounds cull correctly with no GC spikes.

### DIFF
--- a/Assets/Gothic-Core/Scripts/Domain/Culling/VobMeshCullingDomain.cs
+++ b/Assets/Gothic-Core/Scripts/Domain/Culling/VobMeshCullingDomain.cs
@@ -1,11 +1,11 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Linq;
 using Gothic.Core.Adapters.Vob;
 using Gothic.Core.Debugging;
 using Gothic.Core.Logging;
 using Gothic.Core.Manager;
+using Gothic.Core.Models.Config;
 using Gothic.Core.Models.Container;
 using Gothic.Core.Services.Caches;
 using Gothic.Core.Services.Context;
@@ -14,9 +14,9 @@ using Gothic.Core.Extensions;
 using MyBox;
 using Reflex.Attributes;
 using UnityEngine;
-using ZenKit;
 using ZenKit.Vobs;
 using Logger = Gothic.Core.Logging.Logger;
+using NumericsVector3 = System.Numerics.Vector3;
 
 namespace Gothic.Core.Domain.Culling
 {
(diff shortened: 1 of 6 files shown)


Assets/Gothic-Core/Scripts/Domain/Culling/VobMeshCullingDomain.cs
```diff
@@ -80,250 +108,260 @@ public override void Init()
...
-            VobChanged(evt, _objectsSmall, VobList.Small);
         }
 
-        private void VobMediumChanged(CullingGroupEvent evt)
+        /// <summary>
+        /// Band 0 - [0...cullingDistance)         - VOB is enabled.
+        /// Band 1 - [cullingDistance...*1.15)     - Hysteresis zone: VOB keeps its current state to avoid border flickering.
+        /// Band 2 - [cullingDistance*1.15...inf)  - VOB is disabled.
+        ///
+        /// We deliberately ignore evt.isVisible: Frustum culling of Renderers is done by Unity/URP itself.
```
JaXt0r, line 206: I don't know if that is correct. We need to check in Scene View when game is running. Turning around and see, if a Vob is getting invisible or stays. Because Frustum should be handled for static objects. But how about Items?

JaXt0r: I tested it. Works as expected. Thanks a lot. Less checks per frame are always appreciated. :-)

JaXt0r (approved):
Works like a charm. Performance improvements like this are always appreciated. 🙇

Assets/Gothic-Core/Scripts/Domain/Culling/VobMeshCullingDomain.cs
JucanAndreiDaniel: No more custom frustum, everything that is in range is left active but culled by unity itself. in scene view we can still see the vobs, but that's because the scene view is another camera.

But perhaps there is a case for truly static vobs (not interactable/movable/items) to be culled by custom frustum.

PR #373, opened 2026-06-15
Stack/06 texture cache

JucanAndreiDaniel:
## Summary
Refactor the texture cache to allow a material to be in different texture cache arrays (OWODWAT_A0 is used by river materials and by the Old Camp cauldron's soup surface)

## How to test

Load into world.zen and check river in front of old camp

### DIFF
--- a/Assets/Gothic-Core/Scripts/Domain/StaticCache/TextureArrayCacheCreatorDomain.cs
+++ b/Assets/Gothic-Core/Scripts/Domain/StaticCache/TextureArrayCacheCreatorDomain.cs
@@ -23,7 +23,17 @@ namespace Gothic.Core.Domain.StaticCache
 {
     public class TextureArrayCacheCreatorDomain
     {
-        public Dictionary<string, StaticCacheService.TextureInfo> TextureArrayInformation { get; } = new();
+        /// <summary>
+        /// Texture information per target array. A texture can be registered in the Water array AND a solid
+        /// array at the same time (dual-use, e.g. G1's OWODWAT_A0 is used by river materials and by the Old
+        /// Camp cauldron's soup surface). Gothic's VFS is case-insensitive, therefore the keys are as well.
+        /// </summary>
+        public Dictionary<TextureCacheService.TextureArrayTypes, Dictionary<string, StaticCacheService.TextureInfo>> TextureArrayInformation { get; } = new()
+        {
+            [TextureCacheService.TextureArrayTypes.Opaque] = new Dictionary<string, StaticCacheService.TextureInfo>(StringComparer.OrdinalIgnoreCase),
+            [TextureCacheService.TextureArrayTypes.Transparent] = new Dictionary<string, StaticCacheService.TextureInfo>(StringComparer.OrdinalIgnoreCase),
+            [TextureCacheService.TextureArrayTypes.Water] = new Dictionary<string, StaticCacheService.TextureInfo>(StringComparer.OrdinalIgnoreCase)
+        };
 
         [Inject] private readonly VmCacheService _vmCacheService;
         [Inject] private readonly FrameSkipperService _frameSkipperService;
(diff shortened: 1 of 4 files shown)


Assets/Gothic-Core/Scripts/Domain/StaticCache/TextureArrayCacheCreatorDomain.cs
```diff
@@ -200,8 +205,19 @@ private void AddTexInfoForSingleVob(IVirtualObject vob)
 
         private void AddTextureToCache(MaterialGroup group, string textureName)
         {
-            // Already cached.
-            if (TextureArrayInformation.ContainsKey(textureName))
+            // The target array is a pure function of (material group, texture format): registration is
+            // deterministic and independent of the order in which worlds, VOBs and items are processed.
+            // Dual-use textures get registered once per referencing side (Water and solid).
+            if (group == MaterialGroup.Water)
```
JaXt0r, line 211: I don't think this separate IF is needed. Just put it all together:
IF (Water.contains() || Opaque.Contains() || Transparent.Contains()
Helps better readability.

Assets/Gothic-Core/Scripts/Domain/StaticCache/TextureArrayCacheCreatorDomain.cs
```diff
@@ -220,40 +236,39 @@ private void AddTextureToCache(MaterialGroup group, string textureName)
...
-            TextureArrayInformation.TryAdd(textureName,
+            textures.TryAdd(textureName,
                 new StaticCacheService.TextureInfo(textureArrayType, Math.Max(texture.Width, texture.Height), animationTextures.Count));
 
             // If the texture is an "animated one", we also need to add the animation textures. During runtime, water will iterate the z-index of TextureArray to loop through these elements.
             foreach (var animationTexture in animationTextures)
             {
-                TextureArrayInformation.Add(animationTexture.Key,
-                    new StaticCacheService.TextureInfo(textureArrayType, Math.Max(animationTexture.Value.Width, animationTexture.Value.Height), 0));
+                // Animation frames must occupy the array slices directly after their base texture.
```
JaXt0r, line 264: Please add more words from your end why this is needed as comment in code. I currently don't understand.

Assets/Gothic-Core/Scripts/Services/Caches/TextureCacheService.cs
```diff
@@ -162,34 +162,26 @@ public Texture2D TryGetTexture(ITexture zkTexture, string key, bool useCache = t
         public void GetTextureArrayIndex(IMaterial materialData, out TextureArrayTypes textureArrayType, out int arrayIndex, out Vector2 textureScale, out int maxMipLevel, out int animFrameCount)
         {
             var textureName = materialData.Texture;
-            var texture = _resourceCacheService.TryGetTexture(textureName);
+            (int Index, StaticCacheService.TextureInfo Data) entry;
 
-            if (_staticCacheService.LoadedTextureInfoOpaque.ContainsKey(materialData.Texture))
+            // A texture can live in the Water array AND a solid array at the same time (dual-use, e.g. G1's
```
JaXt0r, line 167: Ah. That's the fix!
This comment is also worth a lot for later understandings.

JaXt0r (commented):
Just a few comments.

Assets/Gothic-Core/Scripts/Domain/StaticCache/TextureArrayCacheCreatorDomain.cs
JucanAndreiDaniel: this comment is to be taken in context with the one from the foreach loop

Assets/Gothic-Core/Scripts/Domain/StaticCache/TextureArrayCacheCreatorDomain.cs
JucanAndreiDaniel: now that i think about it the whole if/else if logic is not robust enough, this only checks if a texture with the water material is used in water and other caches. dual-textures can only be in water and either opaque or transparent because of how they are processed later in the function

i need to take a look at this

JaXt0r (approved):

PR #376, opened 2026-06-15
feat: move culled NPCs to their routine waypoint on schedule change

BoroBongo:
NpcRoutineService.CalculateCurrentRoutine now calls NotifyNpcRoutineChanged when the active routine slot changes, moving the culling sphere (and unloaded GO) to the new waypoint so off-screen NPCs progress their daily schedule. RecalculateAllNpcRoutines() iterates every NPC container and can be called after a time skip to synchronise all culled NPCs at once.

### Summary
- When a routine schedule change fires (e.g. time of day changes and an NPC should move to a new waypoint), NPCs that are currently culled (outside the active culling distance) are now teleported to their new routine waypoint immediately
- Previously culled NPCs would stay at their last known position and snap back weirdly only when the player got close enough to un-cull them
- `NpcMeshCullingDomain` and `NpcMeshCullingService` coordinate with `NpcRoutineService` to apply the waypoint update without enabling the NPC's full GameObject

### How to test
1. Find an NPC with a time-based routine (e.g. a guard who walks to a fire at night) - I tested it with Kalom resetting routines after giving him Almanach and triggering THE GREAT SUMMONING
2. Walk far enough away that the NPC culls (disappears from active scene)
3. Use the time skip cheat (or wait) until the NPC's routine schedule changes
4. Walk back to the NPC's **new** routine waypoint — verify the NPC is there, not at their old position
5. Repeat with a sleeping NPC whose morning routine moves them to a work spot


Assets/Gothic-Core/Scripts/Services/Npc/NpcRoutineService.cs
```diff
@@ -140,7 +140,25 @@ public bool CalculateCurrentRoutine(NpcInstance npc)
...
+
             return changed;
         }
+
+        /// <summary>
+        /// Re-evaluates the time-based routine for all NPCs. Culled NPCs whose routine changed
+        /// will have their culling sphere moved to the new waypoint via NotifyNpcRoutineChanged.
+        /// Call this after a time skip so off-screen NPCs appear at their correct schedule position.
+        /// </summary>
+        public void RecalculateAllNpcRoutines()
```
JaXt0r, line 154: This function is never called. Remove if unused (Especially as it can be a performance bottleneck).

Assets/Gothic-Core/Scripts/Domain/Culling/NpcMeshCullingDomain.cs
```diff
@@ -253,5 +253,11 @@ public List<NpcContainer> GetVisibleNpcs()
 
             return visibleNpcs;
         }
+
+        public IEnumerable<NpcContainer> GetAllNpcContainers()
```
JaXt0r, line 257: This function is never called. Remove if unused (Especially as it can be a performance bottleneck).

Assets/Gothic-Core/Scripts/Services/Culling/NpcMeshCullingService.cs
```diff
@@ -44,5 +44,10 @@ public List<NpcContainer> GetVisibleNpcs()
         {
             return _npcDomain.GetVisibleNpcs();
         }
+
+        public IEnumerable<NpcContainer> GetAllNpcContainers()
```
JaXt0r, line 48: This function is never called. Remove if unused (Especially as it can be a performance bottleneck).

Assets/Gothic-Core/Scripts/Services/Npc/NpcRoutineService.cs
```diff
@@ -140,7 +140,25 @@ public bool CalculateCurrentRoutine(NpcInstance npc)
             }
             npcProps.RoutineCurrent = newRoutine;
 
+            if (changed)
```
JaXt0r, line 143: Missing line of code. Good finding. :-)

JaXt0r (changes requested):

PR #377, opened 2026-06-15
Feat/marvin cheats

BoroBongo:
This one will work best once PR Status Menu #378 will be implemented.

To test:
- Tick Marvin Mode and all of the cheats in DevConfig
- Start game
- Open StatusMenu
- On Marvin menu - click buttons

You should see fields being updated.


JaXt0r (changes requested):
Marvin mode is already implemented at your left hand. Just check for Marvin* files. It is one developer config. Add more features you want there please.

JaXt0r:
> Marvin mode is already implemented at your left hand. Just check for Marvin* files. It is one developer config. Add more features you want there please.

Ah I see. At least you can remove the DevsConfig settings. Simply put your new Marvin entries there. No need for them as separate settings I guess?

PR #378, opened 2026-06-15
feat: StatusMenu live hero stats and guild name display

BoroBongo:
StatusMenu was previously a stub with no data. It now reads all hero stats from NpcContainer on open and displays them correctly:

- Guild name via new VmService.GetGuildName() reading TXT_GUILDS from Daedalus (localized name, e.g. "Nowicjusz"), with Props.TrueGuild override support
- Level / XP / LP from hero.Vob (synced from NpcInstance via new NpcService.SyncHeroInstanceToVob() after any VM call)
- HP / Mana / STR / DEX via vob.GetAttribute() index mapping
- Armor (blunt, point, fire, magic) via vob.GetProtection()
- Talents: all skill names and values from VmService.TalentTitles / TalentSkills with pipe-delimited skill-level formatting

QuestLogMenu: BACK arrow was positioned outside the content area; fixed by accounting for localScale.x/y in the offset calculation.

'Nameless'
image

'Diego' 
image

'Milten'
image

'QuestLog'
image


Assets/Gothic-Core/Scripts/Services/Vm/VmService.cs
```diff
@@ -59,6 +59,15 @@ public List<string> TalentSkills
             }
         }
 
+        public string GetGuildName(int guildId)
+        {
+            var guilds = _gameStateService.GothicVm.GetSymbolByName("TXT_GUILDS");
+            var max = _gameStateService.GothicVm.GetSymbolByName("GIL_MAX")?.GetInt(0) ?? 42;
```
JucanAndreiDaniel: where is 42 from?

BoroBongo: There are 42 guilds in code. GIL_MAX is an int = 42 and also 42 is the answer to the universe

JucanAndreiDaniel: is this the same for gothic 2?

BoroBongo: hmmm on gothic 2 its 66 and yeah i didn't think of how mods would behave as well since they will have different number.... I'll get rid of the safety

JucanAndreiDaniel (approved):

PR #372, opened 2026-06-13
Stack/05 npc ai combat

JucanAndreiDaniel:
## Summary
NPC AI, routine, weapon and melee-combat overhaul. Removes the standalone routine service in favour of the NPC services and fills in weapon draw/undraw plus hit detection.

## Changes
- **Routine removal** — deletes `RoutineService`/`RoutineHandler` and their Reflex registration.
- **Routine + AI** — scheduling/progression moved into `NpcRoutineService`, `NpcAiService`, `NpcService`, `NpcHelperService`, wired through `BootstrapService`.
- **Weapons** — `AiDrawWeapon`/`UndrawWeapon` actions using the real `1h`/`2h`/`Mag` animation prefixes, plus the backing VM externals and enums.
- **Combat** — `Attack` hit detection gated on `DEF_OPT_FRAME` with the G1 damage formula; `FightService` clears the queued action list on death so corpses don't play out pending animations; `AiHandler` updates.
- **Action sequencing** — fixes to walk/turn/use-mob actions: `GoToFp` orphan-walk slide, `UseMob` slot tags with stepwise stop, and turn/walk transitions.

## How to test
- Confirm NPCs follow daily routines (walk between waypoints, use mobs/benches).
- Draw and undraw 1h/2h/magic weapons and confirm the correct animations play.
- Engage in melee: confirm hits only register inside the attack frame window and damage matches the G1 formula.
- Kill an NPC mid-action and confirm the corpse doesn't finish queued animations.

## Not fixed
- combat for human NPCs doesn't work

PR #380, opened 2026-06-15
Fix/npc ai fixes

BoroBongo:
### Summary
- Fix NullReferenceException in `ExtNpcGetDistToWp` (`NpcHelperService`) — null check on the NPC GameObject was evaluated *after* accessing `.transform.position`; reordered to guard `!npcGo` first so culled/uninitialized NPCs return `int.MaxValue` cleanly
- Fix null dereference in `UseMob` — the `container!.Go` null-forgiving operator was wrong; replaced with an explicit `container == null` guard so a mob action queued before the container is ready finishes gracefully instead of crashing
- Fix null dereference in `ContinueRoutine` — `PrefabProps.AiHandler` can be null when the action fires on a freshly respawned NPC; added fallback to `GetComponent<AiHandler>()`, logs a warning and early-returns if still null
- Fix stale `CurrentWayPoint`/`CurrentFreePoint` after culling in `AiHandler.ReEnableNpc` — after teleporting an NPC to its routine waypoint on re-enable, the properties were never updated; subsequent `ExtNpcGetDistToWp` calls were operating on stale (or null) position data
- Fix `ExtAiStartState` not immediately clearing the running action when `stopCurrentState = true` (`NpcAiService`) — previously the clear was deferred through the queue; now `ClearState(false)` runs immediately, `StateEnd` is zeroed, and `CurrentWayPoint` is nulled so the new state starts next frame rather than after the whole queue drains
- Fix `BodyState` not reflecting NPC movement and AI transitions — `AbstractWalkAnimationAction2` now sets `BsWalk`/`BsRun` on walk start and `BsStand` on stop; `AiHandler.ClearState` resets to `BsStand`; `ExtAiStandUp` sets `BsStand` immediately (before enqueuing `StandUp`) so Daedalus `Npc_GetBodyState` / `C_BodyStateContains` checks in the same `ZS_*_Loop` tick see the correct value

### How to test
1. Load any outdoor world with NPCs on routines (e.g. Old Camp)
2. Approach an NPC that is walking toward a mob (bench/bed/fire) — verify no crash or freeze when it executes `UseMob`
4. Open a dialog with an NPC mid-walk; verify `Npc_GetBodyState` conditions using `BS_WALK` evaluate correctly (NPC should not appear to be standing while visually walking)
5. Trigger a Daedalus state change via `AI_StartState` with `stopCurrentState = 1` — verify the NPC transitions immediately without waiting for the previous action queue to drain. -- Once you finish talking with Diego, let him take a few steps and then talk to him -> He should now stop and engage in conversation instead of ignoring and focusing on getting to his waypoint.


Assets/Gothic-Core/Scripts/Domain/Npc/Actions/AnimationActions/UseMob.cs
```diff
@@ -49,7 +49,7 @@ public override void Start()
             _mobContainer = container;
             _mobsiScheme = _mobContainer?.Props.GetVisualScheme();
 
-            if (container!.Go == null)
+            if (container == null || container.Go == null)
```
JucanAndreiDaniel, line 52: same fix in #372, we'll have to solve the conflict regardless of which PR gets merged first

JucanAndreiDaniel (approved):

JaXt0r (approved):

PR #382, opened 2026-06-17
fix: properly set the sky textures, only during the day

JucanAndreiDaniel:
## Summary
Sky textures for the day were set during the night

## How to test

Set time to 23 and boot up

### DIFF
--- a/Assets/Gothic-Core/Scripts/Services/World/SkyService.cs
+++ b/Assets/Gothic-Core/Scripts/Services/World/SkyService.cs
@@ -144,21 +144,26 @@ namespace Gothic.Core.Services.World
                 Logger.LogError(e.ToString(), LogCat.Mesh);
                 return;
             }
+            
+            var fogColor = new Vector3(colorValues[0], colorValues[1], colorValues[2]);
 
             foreach (var state in _stateList)
             {
-                // all states that contain day sky layer dawn, evening and day 0 to 3
-                if (state.Time < 0.35 || state.Time > 0.65)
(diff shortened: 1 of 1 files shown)


Assets/Gothic-Core/Scripts/Services/World/SkyService.cs
```diff
@@ -144,21 +144,26 @@ private void UpdateStateTexAndFog()
...
                 return;
             }
+            
+            var fogColor = new Vector3(colorValues[0], colorValues[1], colorValues[2]);
 
             foreach (var state in _stateList)
             {
-                // all states that contain day sky layer dawn, evening and day 0 to 3
-                if (state.Time < 0.35 || state.Time > 0.65)
+                if (!(state.Time < 0.35) || !(state.Time > 0.65)) // set new textures only for day states
```
JaXt0r, line 152: Is time really a value from 0...1? I thought it's an int from 0...23 or so? Can you please check? 0.65 is tough to read instead of e.g. 8am.

JaXt0r: And yes, it was there before, but maybe you find a nicer solution now. 😁

JaXt0r (changes requested):

Assets/Gothic-Core/Scripts/Services/World/SkyService.cs
JucanAndreiDaniel: i want to handle this in a slightly bigger refactor, i don't like it either, this is just a quick fix for the modding weekend

JaXt0r (approved):
Got it. Move on.

